Issue #5, opened 2017-08-06
Field active on rma.operation


andhit-r:
Hi Guys,

IMO, have an active field on master data always useful.

LoisRForgeFlow:
added

Issue #4, opened 2017-08-06
rma_operation_id on product and product.category


andhit-r:
Hi Guys,

Since you have two type of RMA operation (customer and supplier), IMO it will be best if you have 2 rma_operation_id on both product and product.category.

JordiBForgeFlow:
I agree

PR #6, opened 2017-08-16
[9.0][IMP] rma_operation_id on product and product.category

AaronHForgeFlow:
- This solves these issues: https://github.com/Eficent/stock-rma/issues/4 and https://github.com/Eficent/stock-rma/issues/5
- It also separates the rma operation menu between customer and supplier
- Added more improved tests
- Fixing existing previous travis errors
- Various fixes, especially in the wizard that creates supplier rma from a customer rma

### DIFF
--- a/rma/models/product.py
+++ b/rma/models/product.py
@@ -8,7 +8,9 @@ from openerp import fields, models
 class ProductTemplate(models.Model):
     _inherit = 'product.template'
 
-    rma_operation_id = fields.Many2one(
-        comodel_name="rma.operation", string="RMA Operation")
+    rma_customer_operation_id = fields.Many2one(
+        comodel_name="rma.operation", string="Default RMA Customer Operation")
+    rma_supplier_operation_id = fields.Many2one(
+        comodel_name="rma.operation", string="Default RMA Supplier Operation")
     rma_approval_policy = fields.Selection(
         related="categ_id.rma_approval_policy", readonly=True)
--- a/rma/models/rma_operation.py
+++ b/rma/models/rma_operation.py
@@ -26,10 +26,7 @@ class RmaOperation(models.Model):
 
     name = fields.Char('Description', required=True)
     code = fields.Char('Code', required=True)
-    refund_policy = fields.Selection([
-        ('no', 'No refund'), ('ordered', 'Based on Ordered Quantities'),
-        ('received', 'Based on Received Quantities')], string="Refund Policy",
-        default='no')
+    active = fields.Boolean(string='Active', default=True)
     receipt_policy = fields.Selection([
         ('no', 'Not required'), ('ordered', 'Based on Ordered Quantities'),
         ('received', 'Based on Delivered Quantities')],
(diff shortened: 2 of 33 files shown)


AaronHForgeFlow:
@lreficent This is ready to be merge :dancer:

rma/models/rma_order.py
```diff
@@ -175,8 +179,8 @@ def action_rma_done(self):
     def _get_valid_lines(self):
         """:return: A recordset of rma lines.
         """
-        self.ensure_one()
-        return self.rma_line_ids
+        for rec in self:
+            return rec.rma_line_ids
```
LoisRForgeFlow, line 183: Why this? I prefer the approach with the `ensure_one` if you call this method with a recordset a error must raise. If not, you can be disregarding errors.

rma/models/rma_operation.py
```diff
@@ -26,10 +26,7 @@ def _default_supplier_location_id(self):
 
     name = fields.Char('Description', required=True)
     code = fields.Char('Code', required=True)
-    refund_policy = fields.Selection([
```
LoisRForgeFlow, line 29: Why is the refund policy removed?

LoisRForgeFlow: Nevermind, I saw it is moved to rma_account

rma/models/product.py
```diff
@@ -8,7 +8,9 @@
 class ProductTemplate(models.Model):
     _inherit = 'product.template'
 
-    rma_operation_id = fields.Many2one(
```
LoisRForgeFlow, line 11: In the module 'rma_sale'  there are still some references to this field.

rma_purchase/wizards/rma_add_purchase.py
```diff
@@ -75,8 +75,8 @@ def _prepare_rma_line_from_po_line(self, line):
             {'in_route_id': operation.in_route_id.id or route,
              'out_route_id': operation.out_route_id.id or route,
              'receipt_policy': operation.receipt_policy,
-             'location_id': operation.location_id.id or \
-                            self.env.ref('stock.stock_location_stock').id,
+             'location_id': operation.location_id.id or
+                self.env.ref('stock.stock_location_stock').id,
```
LoisRForgeFlow, line 79: I don't think using demo data like this is a good idea...

rma_sale/wizards/rma_add_sale.py
```diff
@@ -74,8 +74,8 @@ def _prepare_rma_line_from_sale_order_line(self, line):
             {'in_route_id': operation.in_route_id.id or route,
              'out_route_id': operation.out_route_id.id or route,
              'receipt_policy': operation.receipt_policy,
-             'location_id': operation.location_id.id or
-                            self.env.ref('stock.stock_location_stock').id,
+             'location_id': operation.location_id.id or self.env.ref(
+                 'stock.stock_location_stock').id,
```
LoisRForgeFlow: same

LoisRForgeFlow (commented):
to use demo data in logic as kind of 'default', are you sure that's a good option?

LoisRForgeFlow:
My last commit (1e5995c006a0156a979b05c9ac07abfa8421db38) attend https://github.com/Eficent/stock-rma/issues/7

LoisRForgeFlow:
I will merge this to continue working without messing up things until @aheficent is available to answer my comments

rma/models/product.py
LoisRForgeFlow: fixed in 1e5995c006a0156a979b05c9ac07abfa8421db38

rma_purchase/wizards/rma_add_purchase.py
AaronHForgeFlow: I agree, remove that

PR #12, opened 2017-10-18
[9.0][REW] rma: workflow centralized on rma.order.line 

LoisRForgeFlow:
changes:
* workflow centralized on rma.order.line 
* the use of rma.order is optional.

pending to test in rma:
* [x] : workflow as group
* [x] : add from stock.move wizard
* [x] : use constrains on customer and lines replacing the state.
* [ ] : tests are failing.

pending to test on rma.line:
* [ ] : lots, select lots in rma.line (better compatibility with v11 stock.move.line)
* [x] : create supplier rma wizard - to be fixed


AaronHForgeFlow:
Please change the version number.
.
.
.
Just kidding :stuck_out_tongue:

PR #16, opened 2017-12-20
[IMP] [9.0] Improved_rma_tests

nikul-serpentcs:
* Improved Stock RMA Tests.

@aheficent


LoisRForgeFlow:
Please rebase

LoisRForgeFlow:
@nikul-serpentcs This is starting to look good! Could you squash the commits?

nikul-serpentcs:
@lreficent Squashed commits.

rma/tests/test_rma.py
```diff
@@ -92,17 +111,54 @@ def _create_rma_from_move(self, products2move, type, partner, dropship,
...
+
+                wizard = self.rma_add_stock_move.with_context(
+                    {'stock_move_id': move.id, 'supplier': True,
+                     'active_ids': [],
+                     'active_model': 'rma.order',
+                     }
+                ).create({})
+                wizard.add_lines()
+
+                wizard = self.rma_add_stock_move.with_context(
```
aheficent: Is this duplicated on purpose?

AaronHForgeFlow, line 124: Is this duplicated on purpose?

rma/tests/test_supplier_rma.py
```diff
@@ -80,12 +76,12 @@ def test_supplier_rma(self):
                 self.assertEquals(line.qty_to_receive, 3,
                                   "Wrong qty to receive")
             if line.product_id == self.product_2:
-                self.assertEquals(line.qty_delivered, 5,
+                self.assertEquals(line.qty_outgoing, 5,
```
AaronHForgeFlow, line 79: So if the transfer is done (line 67) the qty should be already delivered, right?

rma/tests/test_supplier_rma.py
```diff
@@ -100,58 +96,52 @@ def test_supplier_rma(self):
...
-                              "Wrong qty outgoing")
+            self.assertEquals(line.qty_incoming, 0,
+                              "Wrong qty incoming")
             self.assertEquals(line.qty_received, 0,
                               "Wrong qty received")
             if line.product_id == self.product_1:
                 self.assertEquals(line.qty_to_receive, 3,
                                   "Wrong qty to receive")
-                self.assertEquals(line.qty_incoming, 3,
+                self.assertEquals(line.qty_incoming, 0,
```
AaronHForgeFlow, line 113: If the picking is created, the module consider that qty as incoming, so the quantity incoming should be 3 IMHO. If it's zero we will need to check the code.

rma/tests/test_supplier_rma.py
```diff
@@ -100,58 +96,52 @@ def test_supplier_rma(self):
...
-                                  "Wrong qty incoming")
-                self.assertEquals(line.qty_delivered, 2,
-                                  "Wrong qty delivered")
+                self.assertEquals(line.qty_to_deliver, 2,
+                                  "Wrong qty to deliver")
         picking_out.action_assign()
         picking_out.do_transfer()
-        for line in self.rma_supplier_id.rma_line_ids:
-            self.assertEquals(line.qty_to_receive, 0,
+        for line in self.rma_supplier_id.rma_line_ids[0]:
```
AaronHForgeFlow, line 129: Why checking only the first line?

rma/tests/test_supplier_rma.py
```diff
@@ -100,58 +96,52 @@ def test_supplier_rma(self):
...
+            self.assertEquals(line.qty_to_receive, 3,
                               "Wrong qty to receive")
             self.assertEquals(line.qty_incoming, 0,
                               "Wrong qty incoming")
             self.assertEquals(line.qty_to_deliver, 0,
                               "Wrong qty to deliver")
-            self.assertEquals(line.qty_outgoing, 0,
+            self.assertEquals(line.qty_outgoing, 3,
```
AaronHForgeFlow, line 136: If the picking is done (line 128) then the qty_outgoing should be zero but the quantity delivered 3, right?

rma_account/tests/test_rma.py
```diff
@@ -0,0 +1,395 @@
...
+
+    def _create_picking(self, partner):
+        return self.stockpicking.create({
+            'partner_id': partner.id,
+            'picking_type_id': self.env.ref('stock.picking_type_in').id,
+            'location_id': self.stock_location.id,
+            'location_dest_id': self.supplier_location.id
+            })
+
+    def _create_rma_from_move(self, products2move, type, partner, dropship,
```
AaronHForgeFlow, line 70: I'm not sure why this is tested in all the modules, IMHO it's necessary just once, in RMA module

AaronHForgeFlow (commented):
Thanks. It looks ok overall. Just a few comments.

rma/tests/test_rma.py
nikul-serpentcs: @aheficent Because of This wizard in checked operation & route is False and called Wizard method. [here](https://github.com/Eficent/stock-rma/blob/9.0-remove-headers/rma/wizards/rma_add_stock_move.py#L50)

AaronHForgeFlow: ok

AaronHForgeFlow:
Added label because we need to check why the qty is showing outgoing instead of delivered when the picking is confirmed, as shown in my previous comment

rma/tests/test_rma.py
```diff
@@ -111,9 +167,70 @@ def _prepare_move(self, product, qty, src, dest):
...
+            line._default_location_id()
+            line.with_context({'partner_id': line.rma_id.partner_id.id
+                               })._default_delivery_address()
+            line._compute_in_shipment_count()
+            line._compute_out_shipment_count()
+            line._compute_procurement_count()
+
+            data = {'reference_move_id': line.reference_move_id.id}
+            new_line = self.rma_line.new(data)
+            new_line._onchange_reference_move_id()
```
LoisRForgeFlow, line 202: You should test this more specifically, line is now the main document (RMA) so it can work by just itself. The workflow is the following (which you should test independently):

1. You start with a done move for a customer. 
2. You create a RMA (rma.order.line).
3. You set the customer and select the move that need to go thorugh a RMA process.
4. At this moment from the user experience all the field are auto-filled (due `_onchange_reference_move_id`). You should assert that everything is filled correctly.

Same workflow should be tested for extension modules:

* rma_sale:  fill customer and the `sale_line_id` and the method `_onchange_sale_line_id` will fill everything.
* rma_account: customer and `invoice_line_id` -> `_onchange_invoice_line_id`.
* rma_purchase (RMA supplier): customer and `purchase_order_line_id` -> `_onchange_purchase_order_line_id`.

nikul-serpentcs: Okay, So I will test the onchange conditions in all aspects and will ensure that right values are filled into the particular fields.

Is that we have to achieve now, right?

nikul-serpentcs:
@lreficent @aheficent Could you Please take a look??

LoisRForgeFlow:
@nikul-serpentcs Please, could you solve the branch conflicts?

PR #20, opened 2018-01-03
[9.0][ADD] migration scripts

AaronHForgeFlow:


AaronHForgeFlow:
Closing this, the migration scripts are now in a hook, they will be run even when the old rma was not installed. I'll keep the branch in any case.

PS: Some of the changes here implies little fixes, I'll apply those directly to v9 branch

PR #21, opened 2018-01-04
[9.0-remove-headers][FIX]qty to supplier rma

AaronHForgeFlow:
I found out that if I create a picking form scratch the stock move has no partner. There's a constraint that compares the origin move's partner with the partner in the rma line. This fix prevent the constraint to be raised in case of manual picking.


LoisRForgeFlow (approved):
It makes sense :+1:

AaronHForgeFlow:
Merging as travis errors are being fixed here: #16

PR #33, opened 2018-04-17
[9.0][IMP] rma: usability improvements

LoisRForgeFlow:
See the list of commits for the list of features added.


LoisRForgeFlow:
TODO:

* [x] Fix tests

rma/models/rma_order_line.py
```diff
@@ -86,9 +84,11 @@ def _compute_qty_to_receive(self):
         for rec in self:
             rec.qty_to_receive = 0.0
             if rec.receipt_policy == 'ordered':
-                rec.qty_to_receive = rec.product_qty - rec.qty_received
+                rec.qty_to_receive = \
+                    rec.product_qty - rec.qty_incoming -rec.qty_received
             elif rec.receipt_policy == 'delivered':
-                rec.qty_to_receive = rec.qty_delivered - rec.qty_received
+                rec.qty_to_receive = \
+                    rec.qty_delivered - rec.qty_incoming - rec.qty_received
```
AaronHForgeFlow: We have to do the same in the _compute_qty_to_receive method with the qty_outgoing. I write this comment here as I reminder for myself :smile:

rma/models/rma_order_line.py
```diff
@@ -597,3 +618,26 @@ def action_view_procurements(self):
...
+            res = self.env.ref('rma.view_rma_line_supplier_form', False)
+        else:
+            # from supplier we link to customer rma
+            action = self.env.ref(
+                'rma.action_rma_customer_lines')
+            rma_lines = self.customer_rma_id.ids
+            res = self.env.ref('rma.view_rma_line_form', False)
+        result = action.read()[0]
+        # choose the view_mode accordingly
+        if len(rma_lines) != 1:
```
AaronHForgeFlow, line 643: it fails if the supplier rma lines is not created yet

rma/models/rma_order_line.py
LoisRForgeFlow: done!

rma/models/rma_order_line.py
LoisRForgeFlow: hide the button if not supplier rma?

AaronHForgeFlow: yes, that would be great

LoisRForgeFlow: done, but there was in fact a bug in the method. Fixed.

LoisRForgeFlow:
data with rma operation on rma_sale?

AaronHForgeFlow:
Yes, I think it's fine to include one customer operation with sale_policy on received quantities.

LoisRForgeFlow:
remove requested by

LoisRForgeFlow:
* [x] remove warehouse from dropship rules
* [x] when creating operations set the default route if possible
* [x] default customer and supplier operations fields -> filter by type

lreficent:
when creating operations set the default route if possible

LoisRForgeFlow:
security issue with Stock/manager

aheficent:
https://media.giphy.com/media/xdnytp8742kg0/giphy.gif

lreficent:
default customer and supplier operations fields -> filter by type

AaronHForgeFlow:
In v10 the accessError is given in this line. Probably it's happening in v9 also
https://github.com/Eficent/stock-rma/blob/9.0/rma/wizards/rma_make_picking.py#L146

PR #36, opened 2018-06-25
[FIX] Missing Access Rights

max3903:


jgrandguillaume (approved):
LGMT

bodedra (approved):

guewen (approved):

AaronHForgeFlow:
Hi all,

We have been making some fixings and improvements in v9 at https://github.com/Eficent/stock-rma/tree/9.0-imp-rma and in v10 at https://github.com/Eficent/stock-rma/tree/10.0-imp-rma

Among them, we have made the rma_supplier_user and rma_customer_user to inherit permissions of the stock_user: (https://github.com/Eficent/stock-rma/blob/10.0-imp-rma/rma/security/rma.xml#L11)

In this case, I think it's fine if we take them inherit the permissions of the stock_user because the rma_user will do mostly stock operations. In v11 the stock user has access to:

`model_stock_picking,stock.group_stock_user,1,1,1,1
model_stock_picking_type,stock.group_stock_user,1,0,0,0
model_stock_production_lot,stock.group_stock_user,1,1,1,1
model_stock_move,stock.group_stock_user,1,1,1,0
model_stock_inventory,stock.group_stock_user,1,1,1,0
model_stock_inventory_line,stock.group_stock_user,1,1,1,0
product.model_product_product,stock.group_stock_user,1,0,0,0
product.model_product_template,stock.group_stock_user,1,0,0,0
model_stock_warehouse_orderpoint,stock.group_stock_user,1,0,0,0
model_stock_quant,stock.group_stock_user,1,0,0,0
model_stock_quant_package,stock.group_stock_user,1,1,1,1
model_stock_location_path,stock.group_stock_user,1,0,0,0
model_stock_location_path,stock.group_stock_user,1,1,1,1
model_stock_move_line,stock.group_stock_user,1,1,1,1
model_stock_fixed_putaway_strat,stock.group_stock_user,1,0,0,0
product.model_product_price_history,stock.group_stock_user,1,0,0,0
barcodes.model_barcode_nomenclature,stock.group_stock_user,1,0,0,0
barcodes.model_barcode_rule,stock.group_stock_user,1,0,0,0
model_stock_scrap,stock.group_stock_user,1,1,1,0`

IMHO I think giving those permissions to the rma_user is ok. Potentially they need read access to locations, quants, scrap, or picking types, so making rma_user to inherit the permissions of the stock user will solve that problem. I don't think those are too many permissions as the stock_user has only few delete permissions. What do you think @max3903  ?

I will create an issue in this repo so we all can share the latest changes in the different versions (v9,v10 and v11). I will ping you all. It is best to don't have differences between versions so we:
- avoid duplicated efforts
- the stock-rma works the same for all version 

Regards

max3903:
@aheficent I liked the way it was because it fits organizations that have separate teams for warehouse management and RMA, and for organizations that have one team, it's just a dependency to set on a group.

jgrandguillaume:
@aheficent Hi,

After discussing this with @max3903, we suggest the following:

* Maxime will changes this PR to
  * Include the inheritance of the stock_user on both supplier and customer rma groups
  * Sill add the permission on base.user as follow (because this was the main issue on our side, preventing from employee to create new contact):
`
access_rma_order_line_user,access_rma_order_line,model_rma_order_line,base.group_user,1,0,0,0
`
 * You agree to merge this PR as it is consistent with other version

Deal :) ?

Side note on four file "https://github.com/Eficent/stock-rma/blob/10.0-imp-rma/rma/security/rma.xml", you don't need to define two times the inheritance. If your define the inheritance here (https://github.com/Eficent/stock-rma/blob/2daa98e188ea01e4130894369d997e55fc076e08/rma/security/rma.xml#L9) and here (https://github.com/Eficent/stock-rma/blob/2daa98e188ea01e4130894369d997e55fc076e08/rma/security/rma.xml#L15)
You don't need to define it there (https://github.com/Eficent/stock-rma/blob/2daa98e188ea01e4130894369d997e55fc076e08/rma/security/rma.xml#L42)


Regards,

Joël

AaronHForgeFlow:
@jgrandguillaume All right, deal. That would be consistent with all versions. I'll backport  your changes to v10 and v9 then. Thank you!

jgrandguillaume:
@aheficent  Great ! Thanks a lot ! I'll let you merge it when you're ok.

AaronHForgeFlow:
Hi @jgrandguillaume I've tested and I cannot see what is your issue:

without commit d223f15ea3d73269532ae3ed813b68be8387222b a employee has no access to rma_lines but he can create contacts

With that commit he has access to the rma_lines and create contacts.

So what I understand here, the purpose of this is to give access to employees to rma lines but not to the other stock user rights, is it that?

max3903:
@aheficent An employee was not able to create a partner without read access to rma.order.line.

AaronHForgeFlow:
@max3903 If the employee has no access to the rma line he does not have the chance to go there and create the partner. Do you mean to create a partner for a different document?

Not sure, I tested with a user without rma permissions, just employee, and that user was able to create partners. I don't see the relationship here. Am I wrong?

max3903:
@aheficent I will have to test again, cause we don't get the same result...

leemannd:
@max3903 Did you had the time to test it?

max3903:
@aheficent, @leemannd I installed a new DB with Contacts and RMA and the following test user:
screenshot from 2018-07-09 08-52-20

I get the following message when creating a partner:
```
Sorry, you are not allowed to access this document. Only users with the following access level are currently allowed to do that:
- RMA/RMA Customer User
- RMA/RMA Supplier User
- RMA/RMA Manager

(Document model: rma.order.line)
```

AaronHForgeFlow (approved):
@max3903 I guess this is due the compute methods in the res_partner. I think it's ok to add the rules this way. Thank you! :+1:

PR #31, opened 2018-02-09
[11.0] MIG: rma_purchase

bodedra:

PR #39, opened 2018-07-19
[FIX] TypeError: unhashable type: 'list'

bodedra:
Review:
-----------

https://github.com/Eficent/stock-rma/pull/31#issuecomment-400986745

### DIFF
--- a/rma/models/rma_order.py
+++ b/rma/models/rma_order.py
@@ -109,9 +109,9 @@ class RmaOrder(models.Model):
                 for move in line.move_ids:
                     if move.picking_id.location_id == suppliers:
                         picking_ids.append(move.picking_id.id)
-        shipments = list(set(picking_ids))
-        # choose the view_mode accordingly
-        if shipments:
+        if picking_ids:
+            shipments = list(set(picking_ids))
+            # choose the view_mode accordingly
             if len(shipments) > 1:
                 result['domain'] = [('id', 'in', shipments)]
             else:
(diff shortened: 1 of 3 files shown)


AaronHForgeFlow:
Hi @bodedra I'm still getting an error when putting the route "Buy" in incoming shipments. But anyway you can ignore that. As we are including a purchase policy in rma _purchase module we can consider the route buy as a configuration error in a implementation. We can merge this as is not affecting the normal behavior and this is preventing to get an error when no shipment is created in the wizards. Thanks

AaronHForgeFlow (approved):
:+1:

bodedra:
@aheficent yes I am agree with you. When we create incoming shipment with Buy. Odoo creates a purchase order. And our code try to fetch view of incoming picking. That's lead error.

In this case, do we want to create incoming shipment or redirect to purchase order form or any new way?

AaronHForgeFlow:
@bodedra for now, let's consider it a configuration error, so let's keep it like that. As users con create PO directly from a specific button, it makes not much sense to create those form an incoming shipment.

Regards.

AaronHForgeFlow:
@bodedra actually, I was wrong.

The module that check if there's a purchase order related to the rma is the rma always. The thing is, the recommended way to create purchase order from the rma is by using the custom wizard.

However, if the user selects a buy route and create incoming shipment, then the button should return the purchase order, that is the current behavior  in v10. The code that is supposed to do that is the following:
https://github.com/ursais/stock-rma/blob/7e2724118e0c63d43d0fe61287aa503571339bdd/rma_purchase/wizards/rma_make_picking.py#L20

But for some reason is not working in v11. I'll take a further look at it and try to fix it. To fix it we can use this PR https://github.com/Eficent/stock-rma/pull/31 because this is code of the rma_purchase

PR #44, opened 2018-10-04
[11.0][ADD] rma report

AaronHForgeFlow:
Includes RMA report ported from previous versions.

Fixes currency_id field.


AaronHForgeFlow:
Merging, as include fix and it is tested.

PR #46, opened 2018-11-06
[11.0][MIG] account_move_line_rma_order_line

AaronHForgeFlow:


LoisRForgeFlow (approved):
Looks good!

AaronHForgeFlow:
It was working before, but it was adding the rma_line to all the account move lines. After last commit it will do that only for stock_input or stock_output accounts

PR #55, opened 2018-11-21
[11.0][MIG] rma_repair

AdriaGForgeFlow:
Migration of rma_repair module to v11.0


rma_repair/models/rma_order_line.py
```diff
@@ -75,8 +74,7 @@ def _get_rma_repaired_qty(self):
         qty = 0.0
         for repair in self.repair_ids.filtered(
                 lambda p: p.state != 'cancel'):
-            repair_qty = self.env['product.uom']._compute_qty_obj(
-                self.uom_id,
+            repair_qty = self.env['product.uom']._compute_quantity(
```
LoisRForgeFlow: in latests versions you have to call `_compute_quantity` from the uom you want to convert from. See https://github.com/odoo/odoo/blob/11.0/addons/product/models/product_uom.py#L92. So in this case it would be:
```python
repair_qty = self.uom_id._compute_quantity(
    repair.product_qty,
    repair.product_uom,
 )
```

rma_repair/tests/test_rma_repair.py
```diff
@@ -0,0 +1,155 @@
...
+        """Test RMA quantities using repair operations."""
+        # Received sale_policy:
+        rma_1 = self.rma_group_customer.rma_line_ids.filtered(
+            lambda r: r.product_id == self.product_1)
+        self.assertEquals(rma_1.operation_id.repair_type, 'received',
+                          "Incorrect Repair operation")
+        self.assertEquals(rma_1.qty_to_repair, 0.0,
+                          "Quantity to repair should be 0.0")
+        # TODO: receive and check qty_to_sell is 20.0
+        # Ordered sale_policy:
```
LoisRForgeFlow: probably these comments don't belong here.

rma_repair/tests/test_rma_repair.py
```diff
@@ -0,0 +1,155 @@
...
+                          "Operation should be operation_1")
+        rma_2 = self.rma_group_customer.rma_line_ids.filtered(
+            lambda r: r.product_id == self.product_2)
+        rma_2.repair_type = self.operation_2.repair_type
+        self.assertEquals(rma_2.operation_id, self.operation_2,
+                          "Operation should be operation_2")
+
+    def test_02_rma_repair_operation(self):
+        """Test RMA quantities using repair operations."""
+        # Received sale_policy:
```
LoisRForgeFlow: repair_policy

rma_repair/README.rst
```diff
@@ -1,12 +1,39 @@
...
 ==========
 RMA Repair
 ==========
 
+.. !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
+   !! This file is generated by oca-gen-addon-readme !!
+   !! changes will be overwritten.                   !!
+   !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
+
+.. |badge1| image:: https://img.shields.io/badge/maturity-Beta-yellow.png
```
LoisRForgeFlow: Fragmented readme only applies to OCA repositories, sadly in Eficent readmes need to be manually done. In this case original readme I think is fine, no need to change because the menus and workflow stays unchanged.

LoisRForgeFlow (changes requested):
Looks good! Some comments inline.

LoisRForgeFlow:
Code and tests looks good. functionally I found two extra issues:
1. Could you try to remove the save and discard buttons from the wizard? The other wizards from the rma.order.line doesn't have those buttons.
image

2. When I set thec checkbox "to refurbish" I get this error. Could you check that too?
image

LoisRForgeFlow (approved):
Works nicely now! :+1:

PR #64, opened 2019-03-12
[12.0] rma_account: fix res.groups and test

grindtildeath:
- As I didn't want to add the dependency on `sale`, I did set `base.group_no_one` on `discount` as it is done in `account` module : https://github.com/odoo/odoo/blob/12.0/addons/account/views/account_invoice_view.xml#L53 

- `uom` is already in the dependency chain.

- Tests were failing because invoice was still in draft before validation


AaronHForgeFlow (approved):
LGTM :+1:

PR #73, opened 2019-05-17
[12.0][FIX] rma: Cosmetics + remove auto install

bjeficent:
Cosmetics + remove auto install

### DIFF
--- a/rma/models/rma_order_line.py
+++ b/rma/models/rma_order_line.py
@@ -626,5 +626,5 @@ class RmaOrderLine(models.Model):
     @api.constrains("partner_id", "rma_id")
     def _check_partner_id(self):
         if self.rma_id and self.partner_id != self.rma_id.partner_id:
-            raise UserError(_(
+            raise ValidationError(_(
                 "Group partner and RMA's partner must be the same."))
--- a/rma/models/rma_order_line.py
+++ b/rma/models/rma_order_line.py
@@ -622,3 +622,9 @@ class RmaOrderLine(models.Model):
             result['views'] = [(res and res.id or False, 'form')]
             result['res_id'] = rma_lines[0]
         return result
+
+    @api.constrains("partner_id", "rma_id")
+    def _check_partner_id(self):
+        if self.rma_id and self.partner_id != self.rma_id.partner_id:
+            raise UserError(_(
+                "Group partner and RMA's partner must be the same."))
--- a/rma/README.rst
+++ b/rma/README.rst
@@ -94,6 +94,7 @@ Known issues and Roadmap
 * Dropshipping always counted as a delivery on the smart buttons.
 * Uninstall hook.
 * Constraints instead of required fields on rma.order.line.
+* Rename type field on rma.order and rma.order.line
 
 Bug Tracker
 ===========
--- a/rma/models/rma_order.py
+++ b/rma/models/rma_order.py
@@ -1,8 +1,7 @@
 # Copyright (C) 2017 Eficent Business and IT Consulting Services S.L.
 # License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl.html)
 
-from odoo import api, fields, models, _
-from odoo.exceptions import UserError
+from odoo import api, fields, models
 from datetime import datetime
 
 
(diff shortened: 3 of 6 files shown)


AaronHForgeFlow (approved):
I will not merge yet as there is another problems and I will use this PR to fix them:

- Create separate pickings for the lines even where there is a group (this affects all versions)

- Dropshipment from supplier to customer is not creating the delivery correctly.

rma/models/rma_order_line.py
```diff
@@ -439,7 +439,7 @@ def _onchange_reference_move_id(self):
             return
         if sm.move_line_ids.lot_id:
             if len(sm.move_line_ids.lot_id) > 1:
-                raise UserError(_('To manage lots use RMA groups.'))
+                raise ValidationError(_('To manage lots use RMA groups.'))
```
MiquelRForgeFlow:
```suggestion
                raise UserError(_('To manage lots use RMA groups.'))
```

rma/models/rma_order_line.py
```diff
@@ -478,7 +478,7 @@ def action_rma_to_approve(self):
     @api.multi
     def action_rma_draft(self):
         if self.in_shipment_count or self.out_shipment_count:
-            raise UserError(_(
+            raise ValidationError(_(
```
MiquelRForgeFlow:
```suggestion
            raise UserError(_(
```

rma/models/rma_order_line.py
```diff
@@ -622,3 +622,9 @@ def action_view_rma_lines(self):
             result['views'] = [(res and res.id or False, 'form')]
             result['res_id'] = rma_lines[0]
         return result
+
+    @api.constrains("partner_id", "rma_id")
+    def _check_partner_id(self):
+        if self.rma_id and self.partner_id != self.rma_id.partner_id:
+            raise UserError(_(
```
MiquelRForgeFlow:
```suggestion
            raise ValidationError(_(
```

MiquelRForgeFlow (changes requested):

AaronHForgeFlow:
pleaseapprove

rma/models/rma_order_line.py
MiquelRForgeFlow: why?

rma/models/rma_order_line.py
MiquelRForgeFlow: why?

rma/models/rma_order_line.py
AaronHForgeFlow: Why not?

rma/models/rma_order_line.py
AaronHForgeFlow: why not?

MiquelRForgeFlow: because these errors happen when filling the form. that's why I think they should be UserError, because the user put wrong data or miss to fill something. ValidationError is for constrains, where the system is validating the data when saving/writing.

AaronHForgeFlow: Well, here the user is trying to change the state to draft, the data is correct.

MiquelRForgeFlow: that's why is usererror. if the error of the user trying to change the state. it has nothing to do with the data, thus why is not validationError.

rma/models/rma_order_line.py
MiquelRForgeFlow: this error is in an onchange. should be UserError.

rma/models/rma_order_line.py
AaronHForgeFlow: Ok, thanks for the explanation. Just another question. For the user perspective is there any difference?

AaronHForgeFlow:
All comments attended

rma/models/rma_order_line.py
MiquelRForgeFlow: this is UserError, because is in onchange

MiquelRForgeFlow (changes requested):

MiquelRForgeFlow (approved):
LGTM, although maybe it would be better to change the `type` field by `rma_type` or similar.

AaronHForgeFlow:
> LGTM, although maybe it would be better to change the `type` field by `rma_type` or similar.

:+1: Added to the roadmap

PR #80, opened 2019-06-18
[FIX] multicompany security

gurneyalex:
the rma module had company_id fields on rma.order and rma.order.line but not on rma.operation,
and no global multicompany record rules.

-> we fix this and avoid a mess with people allowed to see an operation in the drop down list
but unable to use it because if points to a warehouse of another company


LoisRForgeFlow (approved):
LGTM :+1:

AaronHForgeFlow (approved):
:+1: Agree. Thanks.

PR #81, opened 2019-06-21
[IMP] rma operation form view

gurneyalex:
use the archive button in the rma.operation form view, rather than
a normal boolean field.


bodedra (approved):
LGTM :+1:

AaronHForgeFlow (approved):
:+1:

PR #82, opened 2019-06-21
[FIX][12.0] multicompany: wrong company on stock rules

gurneyalex:
when working with odoobot account, you would get the wrong company_id
on the RMA stock.rules of the warehouse on which you enabled RMA

-> Use the warehouse's company, and not the user's company


AaronHForgeFlow (approved):
:+1: Fast tracking

PR #84, opened 2019-07-04
[11.0][ENH]Allow to receive RMA parts out of the company to prevent valuation

AaronHForgeFlow:
Companies should be able to choose if they hit accounting when receiving RMA parts:

- The parts  are not part of my company until we decide to move to stock
- Only when moving the parts to actual stock the valuation moves are created


AaronHForgeFlow:
Closed until out  of WIP state

PR #99, opened 2019-09-10
[12.0] RMA Account: use Unit Price from RMA Order

dreispt:
Checking the Accounting entries for the Incoming Shipping, I noticed that the value used was the Product's Cost.
In case of an RMA we want the value to the the Price set on the RMA Order.

### DIFF
--- a/rma_account/models/__init__.py
+++ b/rma_account/models/__init__.py
@@ -5,3 +5,4 @@ from . import rma_order
 from . import rma_order_line
 from . import rma_operation
 from . import invoice
+from . import stock_move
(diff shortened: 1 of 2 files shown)


AaronHForgeFlow (approved):
It is fine to me :+1:

dreispt:
@aheficent Thanks for the quick feedback! Let me make the linter happy, and we're good to go.

AaronHForgeFlow:
@jbeficent do you agree with this?

dreispt:
Hello again, on second though, from a functional PoV this might not make sense: stock move accounting should follow the cost of the goods, even in the return case.

PR #103, opened 2019-09-19
[12.0][ENH]qty under repair on rma

AaronHForgeFlow:
With the current version, as soon as the repair order is created is added as qty repaired. This is not correct. To make it more clear, only when the repair is done it is considered repaired and there is a new field qty_under_repair to be more clear

### DIFF
--- a/rma_repair/__manifest__.py
+++ b/rma_repair/__manifest__.py
@@ -3,7 +3,7 @@
 
 {
     "name": "RMA Repair",
-    "version": "12.0.1.1.0",
+    "version": "12.0.1.2.0",
     "license": "LGPL-3",
     "category": "RMA",
     "summary": "Links RMA with Repairs.",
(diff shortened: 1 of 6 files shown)


bodedra:
@aheficent : This is a nice enhancement. But how you handle extra added new items on repair order ?

AaronHForgeFlow:
> @aheficent : This is a nice enhancement. But how you handle extra added new items on repair order ?

Hi @bodedra, good you like it. I am not sure if I underestand. Do you mean like the following example?
* 1 unit for product A in the RMA
* 2 units for product A in the repair order

bodedra:
@aheficent : I meant with extra new parts in repair order which are different from "Product to Repair"

Firefox_Screenshot_2019-09-19T15-51-33 968Z

AaronHForgeFlow:
@bodedra ok I see what you mean. This is not considered in the RMA. We only consider the product that is being repaired no matter how many components are used in the repair order. From the RMA perspective I think this is fine, because it only cares about the finished product.

bodedra:
@aheficent yes. I am agree. I was curious about out of box. :smile:

AaronHForgeFlow:
Rebased. Merging if pipeline succeeds.

Issue #113, opened 2019-10-29
[10.0]rma_account: searching for invoice lines from the RMA customer, gives nothing


chafique-delli:
If the RMA customer is the primary customer and not the billing contact then the wizard to select invoice lines, finds nothing.
The search domain for invoice lines must use the commercial partner.

AaronHForgeFlow:
The domain now is the following:
      `                     domain="['|',('invoice_id.partner_id', '=', partner_id),('invoice_id.partner_id', 'child_of', partner_id)]"/>`

I will recheck but I think that covers you use case.

chafique-delli:
In fact, I want to talk about the domain for the 'invoice_line_ids' field on the view_rma_add_invoice view of the wizard.
I use the modules of the branch '10.0-imp-rma '
If you want, I can do PR for this branch.

AaronHForgeFlow:
> In fact, I want to talk about the domain for the 'invoice_line_ids' field on the view_rma_add_invoice view of the wizard.
> I use the modules of the branch '10.0-imp-rma '

Hi @chafique-delli perhaps are talking about RMA groups? 

> If you want, I can do PR for this branch.

Sure, if there is a bug better to fix it :) Thank you.

chafique-delli:
Sorry, yes, I want to talk about RMA groups.
You can give me the rights to push on the repository 'stock-rma'.

AaronHForgeFlow:
That is not necessary. You can fork the repository and create a Pull Request to 10.0-imp-rma branch :)

chafique-delli:
OK.

chafique-delli:
https://github.com/Eficent/stock-rma/pull/114

AaronHForgeFlow:
Thanks for fixing this :medal_sports:

PR #114, opened 2019-11-05
[10.0]search invoice lines on the commercial partner

chafique-delli:
To correct issue: https://github.com/Eficent/stock-rma/issues/113

### DIFF
--- a/rma_account/wizards/rma_add_invoice.xml
+++ b/rma_account/wizards/rma_add_invoice.xml
@@ -14,7 +14,7 @@
                 </group>
                 <separator string="Select Invoices lines to add"/>
                 <field name="invoice_line_ids"
-                       domain="[('invoice_id.partner_id', '=', partner_id),
+                       domain="[('invoice_id.commercial_partner_id', '=', partner_id),
                        ('invoice_id.type', '=', 'out_invoice')]">
                     <tree string="Invoice Lines">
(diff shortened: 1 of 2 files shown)


rma_account/wizards/rma_add_invoice.xml
```diff
@@ -14,7 +14,7 @@
                 </group>
                 <separator string="Select Invoices lines to add"/>
                 <field name="invoice_line_ids"
-                       domain="[('invoice_id.partner_id', '=', partner_id),
+                       domain="[('invoice_id.commercial_partner_id', '=', partner_id),
```
AaronHForgeFlow: Thanks for this Pull Request!

Perhaps here we can try here with:

`domain="[('invoice_id.commercial_partner_id', 'child_of', partner_id),`

That way if will take both, the commercial_partner, the partner and all the children. Do you agree?

chafique-delli: Ok, I do that.

AaronHForgeFlow: Thank you. I will test if that woks before merging. Does it work on your side?

chafique-delli: Yes, it works on my side

AaronHForgeFlow (approved):
:+1: LGTM

Issue #112, opened 2019-10-29
[10.0] Why 'auto_install'=True for rma_account, rma_sale, rma_purchase modules ?


chafique-delli:
We do not necessarily want to install all the modules of RMA, that's why, I think that the option 'auto_install' is not necessary.

AaronHForgeFlow:
Yes, I agree, at least for sale and purchases. I think we can change that.

AaronHForgeFlow:
Thanks for the fix :medal_sports:

PR #115, opened 2019-11-05
[10.0][FIX]remove autoinstall for rma_account, rma_sale and rma_purchase modules

chafique-delli:
See issue: https://github.com/Eficent/stock-rma/issues/112


AaronHForgeFlow (approved):
:+1: THanks

PR #121, opened 2019-11-29
[12.0][FIX]rma_sale

AaronHForgeFlow:
Current behavior, very difficult to inherit, need wizard information if we want to change behavior.

+ remove warning because rma_line_id is not a field in the wizard.
+ fix action returned when creating PO from RMA
+ shame


aheficent:
/ocabot merge patch

Merging as obvious error

MiquelRForgeFlow (approved):
LGTM

PR #122, opened 2019-12-06
[10.0][IMP] add report for rma group

chafique-delli:


AaronHForgeFlow:
Hi @chafique-delli Thanks for contributing.

I see that apart from the report you are introducing some new field in the rma group, was that done on purpose?

chafique-delli:
Hi @aheficent ,
The fields I added, they are needed for the report of the rma group.
For the "state" field, I added it because I need to read only the common fields that are in this state in the RMA lines.

AaronHForgeFlow:
I see, code seems correct. Perhaps the compute_state could be optimize but not sure. It seems test fail because of the new constraint added. Could you check?

chafique-delli:
Ok, I check that.

chafique-delli:
If you have an idea for optimizing the compute_state method, I'm interested.

bodedra:
@chafique-delli how about if we first search all records linked with rma_line_ids and then count result based on state. I think that way we can optimize compute_state method.

AaronHForgeFlow:
:+1: Tested locally and works well. The computation is not big deal as long as there are not that many lines in an RMA. 

Thank you!

AaronHForgeFlow (approved):
:+1: good!

AaronHForgeFlow:
Superseded by #124

PR #132, opened 2020-04-06
[12.0][UPT]allow to deliver rmas after repair even when not paid

AaronHForgeFlow:
Currently rmas with delivery policy is based on repair quantity  cannot be delivered until the the repair order is paid. That is too restrictive, better just to inform in the RMA about there's some qty to pay instead of restricting.

TODO:
- [x] qty delivered should not depends on qty_to_pay just on qty_repaired
- [x] allow to invoice from the RMA instead of going to the repair order

cc @AdriaGForgeFlow

### DIFF
--- a/rma_repair/models/rma_order_line.py
+++ b/rma_repair/models/rma_order_line.py
@@ -144,3 +144,8 @@ class RmaOrderLine(models.Model):
         for rec in self.filtered(lambda l: l.delivery_policy == 'repair'):
             rec.qty_to_deliver = rec.qty_repaired - rec.qty_delivered
         return res
+
+    def rma_repair_make_invoice(self):
+        res = self.env['repair.order.make_invoice'].with_context(
+            active_ids=self.repair_ids.ids).create({}).make_invoices()
+        return res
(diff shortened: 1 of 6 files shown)


rma_repair/models/rma_order_line.py
```diff
@@ -142,6 +142,10 @@ def _onchange_operation_id(self):
     def _compute_qty_to_deliver(self):
         res = super(RmaOrderLine, self)._compute_qty_to_deliver()
         for rec in self.filtered(lambda l: l.delivery_policy == 'repair'):
-            rec.qty_to_deliver = rec.qty_repaired - rec.qty_delivered - \
-                rec.qty_to_pay
+            rec.qty_to_deliver = rec.qty_repaired - rec.qty_delivered
+        return res
+
+    def rma_repair_make_invoice(self):
+        res = self.env['repair.order.make_invoice'].with_context(
```
AdriaGForgeFlow, line 149: If there is more than one repair in the RMA, IMO you should group the invoices in one using the `group` field in the `'repair.order.make_invoice'` model.

AaronHForgeFlow: done :)

AdriaGForgeFlow (approved):
LGTM :+1:

AdriaGForgeFlow:
Are we missing something here? Can we merge?

PR #144, opened 2020-12-17
[FIX] ValueError: External ID not found in the system: account.res_partner_action_supplier

bodedra:
Issue:
--------

- Install rma module in plain database. It throws

> ValueError: External ID not found in the system: account.res_partner_action_supplier

This PR will fix value error.


bodedra:
cc @AaronHForgeFlow @LoisRForgeFlow

AaronHForgeFlow:
It seems pre-commit config is not correct. Hope to fix that soon.

PR #131, opened 2020-04-06
[12.0][IMP]Fill rma lines in tree

AaronHForgeFlow:
TODO:
- [x] edit rma lines from the list in the rma group view
- [ ] ~~create rma lines in batch from a list of lots~~

cc @AdriaGForgeFlow

PR #156, opened 2021-01-29
[14.0][IMP]Fill rma lines in tree

MateuGForgeFlow:
Edit rma lines from the list in the rma group view.
cc @AaronHForgeFlow 

Forward port of #131


AaronHForgeFlow:
Great! could you include the same also in the supplier rma view ?

bodedra (approved):
LG :+1:

PR #162, opened 2021-04-16
[14.0][MIG] rma*: ir.actions.act_window has different access

LoisRForgeFlow:
right in v14. Actions that read those records need to use
`sudo` to allow non-admin users to be able to use these actions.

@ForgeFlow


AaronHForgeFlow (approved):
I like to test with admin user :laughing:

PR #176, opened 2021-09-02
[14.0][rma_account][imp] allow to change the refund policy even when the rma has been approved

JordiBForgeFlow:
There's no negative effect in changing the policy, and the good thing is that the company may change their mind after the products have been received into stock.


JordiBForgeFlow:
@MateuGForgeFlow @forgeflow

PR #184, opened 2021-10-08
[14.0][IMP] rma_filter_lot: onchange returned by domain should use ids

LoisRForgeFlow:
* onchange returned by domain should use ids
* rename field to `valid_lot_ids`.
* add small readme

@ForgeFlow


LoisRForgeFlow:
Tests ran locally due to https://github.com/ForgeFlow/stock-rma/issues/185. Local test were good.

PR #191, opened 2021-11-24
[IMP] changed to delivery address and serial number

DBolen1994:


AaronHForgeFlow:
It is ok to me to add the lot but not removing price. Ok to me to put the delivery address. WDYT @ChrisOForgeFlow ?

ChrisOForgeFlow:
> It is ok to me to add the lot but not removing price. Ok to me to put the delivery address. WDYT @ChrisOForgeFlow ?

I agree with you, you should keep the price in the view and add the other field

AaronHForgeFlow:
The serial number was added in this PR: https://github.com/ForgeFlow/stock-rma/pull/340 so I think the changes here are not needed anymore. Sorry for the late review and just reopen the PR if you still think this is needed.

PR #193, opened 2021-12-02
[FIX] rma_repair: Be sure to use records of current company

hbrunn:
in a multi company setting, the tests can fail without this if you happen to get records of the wrong company


JordiBForgeFlow (approved):

JordiBForgeFlow:
@hbrunn thanks for the fix!

hbrunn:
thanks for the repo!

PR #207, opened 2022-01-11
[14.0][ADD] rma_invoice

ChrisOForgeFlow:


rma_invoice/models/rma_order_line.py
```diff
@@ -0,0 +1,76 @@
...
+            action["res_id"] = self.invoice_ids.ids[0]
+        else:
+            domain = safe_eval(action["domain"]) or []
+            domain += [("id", "in", self.invoice_ids.ids)]
+            action["domain"] = action["domain"]
+        return action
+
+    def _prepare_rma_line_from_stock_move(self, sm, lot=False):
+        operation_model = self.env["rma.operation"]
+        res = super()._prepare_rma_line_from_stock_move()
```
AaronHForgeFlow:
```suggestion
        res = super()._prepare_rma_line_from_stock_move(sm, lot)
```

AaronHForgeFlow (commented):
Looks good, a couple of comments:

Once the pre-invoice is created I think the button "create services invoice" should not be highlighted:
image

I am not completely sure what is the best way to archive that, but it would be nice to warn the user that the pre-invoice is not created when the invoice policy is "Create Service Invoice". One option is to change the computation of the fields "qty to Receive", "Qty To Deliver", "Qty To Refund", "Qty To Repair", so they depend on the pre-invoice status. But that requires a lot of development. Another option is to show a warning message I am not sure. @LoisRForgeFlow @ChrisOForgeFlow   WDYT?

PR #212, opened 2022-01-26
[14.0][add] account_mass_reconcile_by_rma_line

ChrisOForgeFlow:


JordiBForgeFlow:
Add account-reconcile in https://github.com/ForgeFlow/stock-rma/blob/14.0/oca_dependencies.txt

ChrisOForgeFlow:
> Add account-reconcile in https://github.com/ForgeFlow/stock-rma/blob/14.0/oca_dependencies.txt

Done!

PR #235, opened 2022-03-03
[14.0][FIX] rma: separate stock.move by rma_line_id to fix picking association

### LINKED ISSUE #217: Batch handling of RMA lines in RMA groups does not generate correct links
Target version: 14.0 (to be ported to other versions if needed)

Steps to reproduce:

1. Create an RMA group for 3 RMA lines (I used the serial number feature for this, but the problem could happen in other scenarios as well - not tested):
image

2. Navigate to lines and approve them (side note, a batch action to request approval and approve would be handy here).
3. Select all 3 lines and create the incoming shipment.
image

4. The incoming shipment will be for 3 units but it will only be linked to the first RMA line.
image

image
image
image

5. Also, if you reserve that incoming shipment, it does not reserve the lots/serials the RMAs were created for.
image

ChrisOForgeFlow:
@ForgeFlow

cc. @LoisRForgeFlow 

Close: #217

### DIFF
--- a/rma/models/stock_move.py
+++ b/rma/models/stock_move.py
@@ -99,3 +99,8 @@ class StockMove(models.Model):
             owner_id=owner_id,
             strict=strict,
         )
+
+    @api.model
+    def _prepare_merge_moves_distinct_fields(self):
+        res = super()._prepare_merge_moves_distinct_fields()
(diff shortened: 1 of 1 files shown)


LoisRForgeFlow (approved):
Tested :+1:

PR #236, opened 2022-03-04
[14.0][IMP] Tests for stock valuation

AaronHForgeFlow:
The price taken is:

- The standard cost when the origin of the rma is unknown
- The value of the layers of the original stock move when the origin is a stock move
- The value of the layers of the stock move associated to the original order (sale/purchase) when the origin is a sale/purchase or invoice

That was fixed in https://github.com/ForgeFlow/stock-rma/pull/231 but now the tests are pending

Test are divided in two:

- [x] Include test that checks the value of the layers created
- [x]  Include test that checks the accounts used

There are also some minor fixes of small issues that arise when doing the tests. This will be backported to 13.0 and forwardported to 15.0

@ForgeFlow

### DIFF
--- a/rma/models/res_partner.py
+++ b/rma/models/res_partner.py
@@ -1,4 +1,4 @@
-# Copyright 2017 ForgeFlow
+# Copyright 2017-22 ForgeFlow
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
 
 from odoo import fields, models
@@ -14,7 +14,10 @@ class ResPartner(models.Model):
     rma_line_ids = fields.One2many(
         comodel_name="rma.order.line", string="RMAs", inverse_name="partner_id"
     )
-    rma_line_count = fields.Integer(compute="_compute_rma_line_count")
+    rma_line_count = fields.Integer(
+        compute="_compute_rma_line_count",
+        compute_sudo=True,
+    )
 
     def action_open_partner_rma(self):
         action = self.env.ref("rma.action_rma_customer_lines")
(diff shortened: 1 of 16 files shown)


rma/models/res_partner.py
```diff
@@ -9,7 +9,7 @@ class ResPartner(models.Model):
 
     def _compute_rma_line_count(self):
         for rec in self:
-            rec.rma_line_count = len(rec.rma_line_ids)
+            rec.sudo().rma_line_count = len(rec.rma_line_ids)
```
LoisRForgeFlow: you can use compute_sudo in the field definition

AaronHForgeFlow: Do you have any idea why the compute_sudo is not working ? :S

AaronHForgeFlow: Hi @LoisRForgeFlow , I am afraid the compute_sudo = True does not work, I am not sure if it is because of the type of field is Integer. It just doesn't work :( So I gave up and put the sudo in the method again. Do you think we can move forward with the PR despite that?

AaronHForgeFlow:
It seems copier update has to be run in the repo but I am having trouble doing that.

rma/models/res_partner.py
AaronHForgeFlow: Finally I got the real issue. compute_sudo works.  The problem was the test calling the _compute_rma_line_count directly with a different user. Calling the compute in the test makes no sense so I removed that call. Now it should be green and ready to be merged.

PR #238, opened 2022-03-04
[15.0][IMP] rma*: Refactor all rma modules in order to consider using the correct price unit in moves

LoisRForgeFlow:
Forward port of https://github.com/ForgeFlow/stock-rma/pull/231.


AaronHForgeFlow:
rebased, no I will apply this: https://github.com/ForgeFlow/stock-rma/pull/236/commits

PR #255, opened 2022-05-18
[14.0][FIX/IMP] Make RMA Operation settings company dependent 

DavidJForgeFlow:


rma/__manifest__.py
```diff
@@ -10,7 +10,7 @@
     "in odoo",
     "author": "ForgeFlow",
     "website": "https://github.com/ForgeFlow/stock-rma",
-    "depends": ["stock", "mail", "web"],
+    "depends": ["stock", "mail", "web", "account"],
```
LoisRForgeFlow, line 13: why the new dependency?

PR #257, opened 2022-05-24
[12.0][FIX] Pass sale order line in stock.rule

astirpe:


LoisRForgeFlow:
@astirpe Will this reduce the quantity delivered in the SO line as a consequence?

astirpe:
@LoisRForgeFlow I didn't pay attention to that, but I can verify it if you didn't already

LoisRForgeFlow:
I did not, I was thinking in possible side effects (also the RMA receipt will show in the SO transfers I think - not verified either).

PR #263, opened 2022-06-08
[14][FIX] count produced products going to customer as out pickings

florian-dacosta:
Fix the case where you produce a product and then send it to the customer. It should count as an outgoing picking in that case.

@LoisRForgeFlow @JordiBForgeFlow

### DIFF
--- a/rma/models/rma_order_line.py
+++ b/rma/models/rma_order_line.py
@@ -64,7 +64,7 @@ class RmaOrderLine(models.Model):
         for move in self.move_ids:
             first_usage = move._get_first_usage()
             last_usage = move._get_last_usage()
-            if first_usage == "internal" and last_usage != "internal":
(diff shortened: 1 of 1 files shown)


florian-dacosta:
Hello @LoisRForgeFlow @JordiBForgeFlow 
Any chance to have a review on this one ?

JordiBForgeFlow:
@DavidJForgeFlow @AaronHForgeFlow can you check?

DavidJForgeFlow (approved):
Functional review :+1:

florian-dacosta:
@AaronHForgeFlow Could you check also se we can get it merged if you are ok too?
I would port that to v15 and 16 then.

AaronHForgeFlow (approved):
Ok for me!

PR #264, opened 2022-06-10
[12.0][IMP] rma: prevent the creation of zero qty moves

StefanRijnhart:
Our users manage to end up with pickings with stock moves that have a zero quantity way too often.


MiquelRForgeFlow (approved):

JordiBForgeFlow (approved):

JordiBForgeFlow:
@StefanRijnhart thanks for the fix. @DavidFIB can you forward port to newer versions?

PR #265, opened 2022-06-15
[15.0][FIX] rma: prevent against warehouse mismatch or missing rules

StefanRijnhart:
When creating pickings, ensure that the applied stock rule was taken from
the operation's routes. Otherwise, the default procurement rules for a
warehouse may kick in, creating incoming customer goods not from the customer
location but from the resupply warehouse.

### DIFF
--- a/rma/models/procurement_group.py
+++ b/rma/models/procurement_group.py
@@ -1,7 +1,8 @@
 # Copyright (C) 2017-22 ForgeFlow S.L.
 # License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl.html)
 
-from odoo import fields, models
+from odoo import _, api, fields, models
+from odoo.exceptions import ValidationError
 
 
 class ProcurementGroup(models.Model):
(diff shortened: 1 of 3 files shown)


JordiBForgeFlow (approved):

CasVissers-360ERP:
@StefanRijnhart 
I get an error when updating the module, an error on my side? 

2022-06-16 07:23:38,040 10 ERROR imt-p-15-0-5152597 odoo.sql_db: bad query: 
        WITH rma_customer_operation_id_field AS (
            SELECT id FROM ir_model_fields WHERE model='product.template'
            AND name='rma_customer_operation_id'
        )
        INSERT INTO ir_property(name, type, fields_id, company_id, res_id,
            value_reference)
        SELECT 'rma_customer_operation_id', 'many2one', rco.id, ro.company_id,
            CONCAT('product.template,', t.id), CONCAT('rma.operation,', ro.id)
        FROM product_template t JOIN rma_operation ro
            ON t.rma_customer_operation_id = ro.id
        , rma_customer_operation_id_field rco
        WHERE ro.company_id IS NOT NULL
              AND NOT EXISTS(SELECT 1
                             FROM ir_property
                             WHERE fields_id=rco.id
                                AND company_id=ro.company_id
                                AND res_id=CONCAT('product.template,', t.id))
    
ERROR: column t.rma_customer_operation_id does not exist
LINE 11:             ON t.rma_customer_operation_id = ro.id
                        ^

StefanRijnhart:
@CasVissers-360ERP that would not be due to this PR. The code that does not work for you is introduced by @LoisRForgeFlow for the migration to 15.0: https://github.com/ForgeFlow/stock-rma/blob/15.0/rma/migrations/15.0.1.1.0/post-migration.py

JordiBForgeFlow:
@CasVissers-360ERP your error is strange. Suggests that the product.template does not have a field 'rma_customer_operation_id', but it exists here: https://github.com/ForgeFlow/stock-rma/blob/15.0/rma/models/product.py#L10

StefanRijnhart:
@CasVissers-360ERP are you actually upgrading from the [OCA `rma` module](https://github.com/oca/rma/tree/14.0/rma)? It is different from this one.

@JordiBForgeFlow thanks for the merge!

CasVissers-360ERP:
I think I had some local issues. Resolved them.

JordiBForgeFlow:
@DavidJForgeFlow can you backport the commit to 14.0?

PR #269, opened 2022-06-16
[FIX]rma: remove test_rma dependency to Account

DavidJForgeFlow:
Removed test_rma.py dependency to Account, 
to do this I have removed the account declarations for the manager user account in the test, and moved the old declaration of rma_manager_user to rma_account_unreconciled/tests, because this tests need the account declaration.

@ForgeFlow


LoisRForgeFlow:
@DavidJForgeFlow Please, do the backport to v13 too.

PR #273, opened 2022-06-23
[14.0] Hide smart button when empty on rma.line

florian-dacosta:
Hello,

I am looking to install these RMA modules for a customer, and one of the main issue I meet, is that there are a lot of useless fields, button, smart button, etc, for the case of this customer.
I guess some feature could be in separated modules, although it is probably a change to do during migration of those modules (and maybe not an easy one), but another way to simplify the view is to hide every thing that is not usefull for the current rma line.

So here, I hide all smart button that are empty, for a start.

If you are ok with the idea, I'll do the same for other stuff. The goal beeing to have a simpler view depending on the option chosen (operation) and the action that have been done.
Example, I have a rma line with "Receipts Policy" = "Not required". I still see the "Create Incoming Shipment" button (which give an error if clicked). I also see information about qty to receive, qty received, etc. 
I guess in this case we should hide all this information.

I'd love a feedback on the idea and if positive I will do other PR around this ! @LoisRForgeFlow

### DIFF
--- a/rma/views/rma_order_line_view.xml
+++ b/rma/views/rma_order_line_view.xml
@@ -100,6 +100,7 @@
                                 name="action_view_in_shipments"
                                 class="oe_stat_button"
                                 icon="fa-truck"
+                                attrs="{'invisible': [('in_shipment_count', '=', 0)]}"
                                 groups="stock.group_stock_user"
                             >
                                 <field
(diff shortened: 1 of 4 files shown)


JordiBForgeFlow (approved):

JordiBForgeFlow:
@florian-dacosta I agree with this PR. Hiding of the smart buttons to prevent creating shipments when the policy does not require it is probably a good idea, now that you can change the policies as required. In general when we designed this module we tried to think in terms of flexibility. We allow the user to do what they want, and just guide her/him without introducing hard restrictions.

JordiBForgeFlow:
@DavidJForgeFlow can you backport to earlier versions, and forward port to 15.0?

PR #275, opened 2022-06-24
[15.0][FIX] rma: improve check on rule selection during procurement

### LINKED ISSUE #274: Use default product routes for delivery
Hello, 
I have an issue with my use case, and it is kind of linked to this recent fix : https://github.com/ForgeFlow/stock-rma/pull/265
I'd like to understand better why there is such a strong will to have mandatory "rma route".

I have multiple customer case where, the company never repair the returned product and always delivery products in RMA cases in the exact same way they would do from a sale order.
 And they have a lot of different route depending on the type of product they ship. So it would be quite heavy (too much configuration) to have to create an rma operation for each type of product, with a exact same route that is configured on the product and then configure these operations on the product side etc.
When just letting odoo chosing the route with no special configuration would work as a charm.

I understand that in other case, we need a specific RMA route and it is important, but I believe it should be optional no?
From what I see, making route optional for now would be quite a lot of change, so I don't  want to go there.

From what I understand from https://github.com/ForgeFlow/stock-rma/pull/265 the goal is to raise an error in case the rma chosen route is not used, due to a configuration error ? (Indeed, the rma route has priority on any other rule so the problem must be only in case of bad configuration right ?)
I wonder if we could find a way that would manage both case (raise in case of bad configuration but leave a way to use routes configured on product/categories/warehouse).

One way could be to allow all rules from any route with the flag `rma_selectable`
Here https://github.com/ForgeFlow/stock-rma/blob/14.0/rma/wizards/rma_make_picking.py#L147
Something like `force_rule_ids = self.env['stock.location.route'].search([('rma_selectable', '=', True)]).rule_ids.ids`
And it would be the exact same allowed rules for delivery which is my actual concern.
I think that would solve both case. 

Anyway, I don't know well these modules yet, so I'd like to have your opinions about this. @StefanRijnhart @JordiBForgeFlow @DavidFIB

StefanRijnhart:
Fixes #274

Thanks to @florian-dacosta for suggesting this approach.

### DIFF
--- a/rma/models/procurement_group.py
+++ b/rma/models/procurement_group.py
@@ -17,21 +17,21 @@ class ProcurementGroup(models.Model):
 
     @api.model
     def _get_rule(self, product_id, location_id, values):
-        """Ensure that the selected rule is from the configured route"""
+        """Ensure that the selected rule is valid for RMAs"""
         res = super()._get_rule(product_id, location_id, values)
-        force_rule_ids = self.env.context.get("rma_force_rule_ids")
-        if force_rule_ids:
-            if res and res.id not in force_rule_ids:
+        rma_route_check = self.env.context.get("rma_route_check")
+        if rma_route_check:
+            if res and not res.route_id.rma_selectable:
                 raise ValidationError(
                     _(
-                        "No rule found in this RMA's configured route for product "
-                        "%(product)s and location %(location)s"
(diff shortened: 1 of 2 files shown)


florian-dacosta (approved):
Code review + test LGTM
Thanks @StefanRijnhart !

JordiBForgeFlow (approved):

PR #283, opened 2022-06-29
[14.0][IMP] Make rma order line view cleaner for user

florian-dacosta:
In this PR, I hide the buttons used to generate document (delivery, reception, refund, etc) if the related policy is set to "no".
When the related policy is "no" and we click on the button, a error is raised to the user anyway.
I also hide the policy related quantity fields when the policy is set to "no".

With this, I think the view is really simpler and the user still has full flexibility since he can change the policy anyway.

It is kind of a complement to https://github.com/ForgeFlow/stock-rma/pull/273

@JordiBForgeFlow


LoisRForgeFlow (approved):
LGTM

PR #286, opened 2022-06-30
[12.0][FIX] Default route in rma.operation

dkubiak789:


AaronHForgeFlow:
Sorry for the late reply, I agree current setup is not multicompany valid. We are going for this change here: https://github.com/ForgeFlow/stock-rma/pull/456 so the routes will be taken from the warehouse. And the warehouse will be the default warehouse of the company. With that I think the issue will be solved.

LoisRForgeFlow:
Closed in favor of discussion in https://github.com/ForgeFlow/stock-rma/pull/456. v12 is no longer mantained.

PR #287, opened 2022-07-02
[14.0][IMP] rma_filter_lot: add possibility to select only the lots issued specifically to the customer of the RMA.

JordiBForgeFlow:
Show only the lots shipped to the particular customer in the RMA. If nothing was shipped to that customer, fall back to showing all the lots available in the customer location.


JordiBForgeFlow:
@DavidFIB can you forward and backport where applicable?

PR #301, opened 2022-10-04
[14.0][FIX] rma_account: get price unit with all related layers on sale moves

ChrisOForgeFlow:
@ForgeFlow

### DIFF
--- a/rma_account/models/procurement.py
+++ b/rma_account/models/procurement.py
@@ -33,8 +33,12 @@ class StockRule(models.Model):
             line = self.env["rma.order.line"].browse([line])
             move = line.reference_move_id
             if move and move.stock_valuation_layer_ids:
-                cost = move.stock_valuation_layer_ids[-1].unit_cost
-                res["price_unit"] = cost
+                layers = move.stock_valuation_layer_ids
+                price_unit = sum(layers.mapped("value")) / sum(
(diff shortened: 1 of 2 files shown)


rma_account/models/procurement.py
```diff
@@ -33,8 +33,12 @@ def _get_stock_move_values(
...
             move = line.reference_move_id
             if move and move.stock_valuation_layer_ids:
-                cost = move.stock_valuation_layer_ids[-1].unit_cost
-                res["price_unit"] = cost
+                layers = move.stock_valuation_layer_ids
+                price_unit = sum(layers.mapped("value")) / sum(
+                    layers.mapped("quantity")
+                )
+
+                res["price_unit"] = price_unit
```
LoisRForgeFlow, line 41: does this make sense for FIFO?

JordiBForgeFlow: @LoisRForgeFlow Makes sense, when you apply landed costs because one move will have the main layer and then subsequent layers with extra cost due to landed costs.

JordiBForgeFlow:
@ChrisOForgeFlow Can you forward port to 15.0?

ChrisOForgeFlow:
> @ChrisOForgeFlow Can you forward port to 15.0?

Sure

PR #302, opened 2022-10-18
[14.0][IMP] rma_delivery: add default_carrier_in_id

mariadforgeflow:
@LoisRForgeFlow

### DIFF
--- a/rma_delivery/migrations/14.0.1.1.0/pre-migration.py
+++ b/rma_delivery/migrations/14.0.1.1.0/pre-migration.py
@@ -1,0 +1,20 @@
+# Copyright 2022 ForgeFlow S.L.
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from openupgradelib import openupgrade
+
+
+@openupgrade.migrate()
+def migrate(env, version):
+    if openupgrade.column_exists(env.cr, "delivery_carrier", "default_carrier_id"):
+        openupgrade.rename_fields(
+            env,
+            [
+                (
+                    "rma.operation",
+                    "rma_operation",
+                    "default_carrier_id",
+                    "default_carrier_out_id",
+                )
+            ],
+        )
--- a/rma_delivery/models/rma_operation.py
+++ b/rma_delivery/models/rma_operation.py
@@ -8,6 +8,16 @@ class RmaOperation(models.Model):
 
     _inherit = "rma.operation"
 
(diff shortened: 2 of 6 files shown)


LoisRForgeFlow:
@ChrisOForgeFlow can you have a look at this one?

rma_delivery/models/rma_operation.py
```diff
@@ -8,6 +8,16 @@ class RmaOperation(models.Model):
 
     _inherit = "rma.operation"
 
-    default_carrier_id = fields.Many2one(
-        comodel_name="delivery.carrier", string="Default carrier", required=False
+    default_carrier_id_in = fields.Many2one(
```
ChrisOForgeFlow: Maybe with change of field name, We should add a migration script to put current value in two new fields

ChrisOForgeFlow (changes requested):
LGTM

rma_delivery/models/rma_operation.py
LoisRForgeFlow: I agree. Also, now that I see it, I would name the fields `default_carrier_in_id` and `default_carrier_out_id` to respect naming [conventions](https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#fields).

ChrisOForgeFlow (approved):
LGTM

rma_delivery/migrations/14.0.1.1.0/pre-migration.py
```diff
@@ -0,0 +1,20 @@
...
+
+@openupgrade.migrate()
+def migrate(env, version):
+    if openupgrade.column_exists(env.cr, "delivery_carrier", "default_carrier_id"):
+        openupgrade.rename_fields(
+            env,
+            [
+                (
+                    "delivery.carrier",
+                    "delivery_carrier",
```
LoisRForgeFlow: This should be the model and table of the field being renamed, in this case `rma.operation`

AaronHForgeFlow:
@LoisRForgeFlow ok to merge?

LoisRForgeFlow:
@AaronHForgeFlow eventually, this wasn't implemented in the project that triggered the need, so up to you to include it or not.

LoisRForgeFlow:
@AaronHForgeFlow I assume that no answer in the last copule of years means that you are not particularly interested in including this PR. Closing.

PR #315, opened 2022-11-23
[14.0][IMP] centralize the logic to get the correct cost of the RMA.

JordiBForgeFlow:
@ForgeFlow


JordiBForgeFlow:
@AaronHForgeFlow Do you agree with this?

AaronHForgeFlow (approved):
Much better like this. Thanks!

PR #334, opened 2022-11-30
[15.0][FIX] fix empty parter on rma picking

JordiBForgeFlow:


LoisRForgeFlow:
Already ported here https://github.com/ForgeFlow/stock-rma/pull/337. Closing.

PR #339, opened 2022-12-01
[16][FIX] rma: Refactor onchange fields to computed fields (partial)

florian-dacosta:
This is still kind of WIP. I think the tests on RMA pass because the Form way to create rma is not really used in the tests.

There is still refactore todo, most of the onchange should be gone...
I am not sure if I will have time to complete these kind of refactores, but feels free to superseed this PR of course.

### DIFF
--- a/rma/models/rma_order_line.py
+++ b/rma/models/rma_order_line.py
@@ -301,6 +301,9 @@ class RmaOrderLine(models.Model):
         string="Unit of Measure",
         required=True,
         readonly=True,
+        compute="_compute_uom_id",
+        precompute=True,
+        store=True,
         states={"draft": [("readonly", False)]},
     )
     price_unit = fields.Monetary(
@@ -377,6 +380,9 @@ class RmaOrderLine(models.Model):
         required=True,
         domain=[("rma_selectable", "=", True)],
         readonly=True,
+        compute="_compute_in_route_id",
+        precompute=True,
+        store=True,
         states={"draft": [("readonly", False)]},
     )
     out_route_id = fields.Many2one(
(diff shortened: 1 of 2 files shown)


LoisRForgeFlow:
@florian-dacosta I just faced some issues that would be solved with this refactor, I may supersede this PR. If you have done more work in this direction, please let me know to not do it twice :)

florian-dacosta:
Hello, nothing more on my side, for me we can merge it already and make additional PR or if you prefer superseed, it is ok for me too of course!

LoisRForgeFlow:
For now I have handled the problem I had with this https://github.com/ForgeFlow/stock-rma/pull/355. My original problem was in v14 and I couldn't take advantage of the precompute yet. The change has been [forward ported as is to v16](https://github.com/ForgeFlow/stock-rma/pull/358), but it would be greate to refactor.

florian-dacosta:
@LoisRForgeFlow I don't think I'll have time to make a refactore of all onchange fields any time soon.
But I think we could remove the WIP tag and merge this PR, and keep accepting PR about these kind of refactore, to make it little by little.
Would it be acceptable ?

rma/models/rma_order_line.py
```diff
@@ -704,7 +721,12 @@ def _onchange_lot_id(self):
         product = self.lot_id.product_id
         if product:
             self.product_id = product
-            self.uom_id = product.uom_id
+
+    @api.depends("uom_id")
+    def _compute_uom_id(self):
+        for rec in self:
+            if rec.product_id:
+                rec.uom_id = rec.product_id.uom_id
```
LoisRForgeFlow: uom_id compute depends on "uom_id"? this doesn't look correct

LoisRForgeFlow (commented):
Hi @florian-dacosta 

I have nothing against merging this, however could first review my comment bellow?

rma/models/rma_order_line.py
florian-dacosta: It is not right indeed, fixed, thanks

LoisRForgeFlow: Thanks! I will fix pre-commit in a different pr if needed.

LoisRForgeFlow (approved):

PR #352, opened 2022-12-27
[16.0][FIX] count produced products going to customer as out pickings

LoisRForgeFlow:
Forward port of #263


florian-dacosta:
thanks for the port

PR #360, opened 2023-01-04
[14.0][FIX] rma_account: reconcile GDNI when receiving but refund is not expected

AaronHForgeFlow:
Tries to fix:
https://github.com/ForgeFlow/stock-rma/issues/356

### DIFF
--- a/rma_account/models/rma_operation.py
+++ b/rma_account/models/rma_operation.py
@@ -28,6 +28,16 @@ class RmaOperation(models.Model):
         comodel_name="account.journal",
         compute="_compute_domain_valid_journal",
     )
+    automated_refund = fields.Boolean(
+        help="In the scenario where a company uses anglo-saxon accounting, if "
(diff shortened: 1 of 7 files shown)


JordiBForgeFlow:
Perhaps the refund wizard should have boolean 'free of charge'.

Then, in the operation you could define that an automated refund, should be created after recepit, for the quantities received. In the operation I would distinguish two booleans:
- Automated refund
- Free of charge refund

AaronHForgeFlow:
Well, I have included the two booleans, but basically, the refund_free_of_charge boolean will always be True when the user marks the bollean "automated_refund". I am not sure if that is going to be used in any other sceneraio, but it is there just in case.

JuanyDForgeflow (approved):

LoisRForgeFlow:
@AaronHForgeFlow could you forward port this?

LoisRForgeFlow:
ping @AaronHForgeFlow

PR #363, opened 2023-01-19
[14.0][FIX] rma_sale, fix performance on search sale order lines

ChrisOForgeFlow:
Change domain to computed field, to allow document can be reprocessed, without this change on reprocess domain lost and search on all sale order lines

### DIFF
--- a/rma_sale/views/rma_order_line_view.xml
+++ b/rma_sale/views/rma_order_line_view.xml
@@ -34,10 +34,16 @@
                 <group name="main_info" position="inside">
                     <field name="sale_id" invisible="1" />
                     <field
+                    name="sale_line_domain_ids"
+                    invisible="1"
+                    widget="many2many_tags"
+                />
+                    <field
                     name="sale_line_id"
                     context="{'rma': True}"
                     options="{'no_create': True}"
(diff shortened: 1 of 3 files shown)


AaronHForgeFlow (approved):
I think it makes sense

rma_sale/views/rma_order_line_view.xml
```diff
@@ -33,11 +33,17 @@
                 </div>
                 <group name="main_info" position="inside">
                     <field name="sale_id" invisible="1" />
+                    <field
+                    name="sale_line_domain_ids"
+                    invisible="1"
+                    widget="many2many_tags"
```
LoisRForgeFlow, line 39: widget in an invisible field?

LoisRForgeFlow (approved):
Looks correct to me!

rma_sale/views/rma_order_line_view.xml
ChrisOForgeFlow: I'm not sure if in this version can works without widget, but in previous version I need to put widget and domain works well

AaronHForgeFlow:
Tested. Works ok. Can you please forward port it?

PR #364, opened 2023-01-23
[14.0][FIX] rma_sale: description not passed to SO

DavidJForgeFlow:
The description wasn't being passed to the SO when changing it at the RMA wizard.

@ForgeFlow


LoisRForgeFlow (approved):
Thanks!

PR #371, opened 2023-02-08
[15.0][FIX] rma_account: reconcile GDNI when receiving but refund is not expected

AaronHForgeFlow:
Forward port of https://github.com/ForgeFlow/stock-rma/pull/360/


AaronHForgeFlow:
Tested locally, merging when green.

PR #373, opened 2023-02-13
[14.0][FIX] rma_sale: duplicate SO also link rma

DavidJForgeFlow:
When duplicating an SO that is linked to an RMA, also links the duplicated one to the original RMA.

@ForgeFlow


LoisRForgeFlow (approved):
Thanks!

PR #393, opened 2023-02-21
[14.0][IMP] rma: limit state statusbar

DavidJForgeFlow:
Limit widget statusbar in rma line to only contain draft, to approve, approved and done.

@ForgeFlow


LoisRForgeFlow (approved):
Thanks

PR #412, opened 2023-03-03
[16.0][FIX] rma: product_uom_qty not in move_line_ids

DavidJForgeFlow:
In Odoo 16.0 "stock.move.line" has no longer field product_uom_qty, now it's reserved_uom_qty.

@ForgeFlow


LoisRForgeFlow (approved):
Indeed, thanks!

PR #413, opened 2023-03-09
[16.0][FIX] rma: multiple fixes

DavidJForgeFlow:
Error in migration, this fixes it. stock.move.line.product_uom_qty to stock.move.line.reserved_uom_qty.

Also corrects the position of field rma_selectable in routes form view.

Deliveries are computed now as a move going out to "customers" or "suppliers".

Assigned to filter now only applied once.

@ForgeFlow

### DIFF
--- a/rma/models/rma_order_line.py
+++ b/rma/models/rma_order_line.py
@@ -77,7 +77,10 @@ class RmaOrderLine(models.Model):
         for move in self.move_ids:
             first_usage = move._get_first_usage()
             last_usage = move._get_last_usage()
-            if first_usage in ("internal", "production") and last_usage != "internal":
+            if first_usage in ("internal", "production") and last_usage in (
+                "customer",
+                "supplier",
+            ):
                 pickings |= move.picking_id
             elif last_usage == "customer" and first_usage == "supplier":
                 pickings |= move.picking_id
@@ -98,12 +101,24 @@ class RmaOrderLine(models.Model):
             product_obj = self.env["uom.uom"]
             qty = 0.0
             if direction == "in":
-                op = ops["="]
-            else:
-                op = ops["!="]
-            for move in rec.move_ids.filtered(
-                lambda m: m.state in states and op(m.location_id.usage, rec.type)
-            ):
+                moves = rec.move_ids.filtered(
+                    lambda m: m.state in states
+                    and (
+                        m.location_id.usage == "supplier"
+                        or m.location_id.usage == "customer"
+                    )
+                    and m.location_dest_id.usage == "internal"
+                )
+            elif direction == "out":
+                moves = rec.move_ids.filtered(
+                    lambda m: m.state in states
+                    and (
+                        m.location_dest_id.usage == "supplier"
+                        or m.location_dest_id.usage == "customer"
+                    )
+                    and m.location_id.usage == "internal"
+                )
+            for move in moves:
                 # If the move is part of a chain don't count it
                 if direction == "out" and move.move_orig_ids:
                     continue
--- a/rma/models/rma_order_line.py
+++ b/rma/models/rma_order_line.py
@@ -68,7 +68,7 @@ class RmaOrderLine(models.Model):
             if last_usage == "internal" and first_usage != "internal":
                 moves |= move
             elif last_usage == "supplier" and first_usage == "customer":
-                moves |= moves
+                moves |= move
         return moves
 
     @api.model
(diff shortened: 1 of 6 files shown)


OriolMForgeFlow (approved):

OriolMForgeFlow (approved):

rma/wizards/rma_make_picking_view.xml
```diff
@@ -108,26 +108,26 @@
...
-                    attrs="{'invisible':['|', '|', ('qty_to_receive', '!=', 0), ('state', '!=', 'approved'), ('receipt_policy', '=', 'no')]}"
+                    attrs="{'invisible':['|', '|', ('qty_to_receive', '>', 0), ('state', '!=', 'approved'), ('receipt_policy', '=', 'no')]}"
                     type="action"
                 />
                 <button
                     name="%(action_rma_picking_out)d"
                     string="Create Delivery"
                     class="oe_highlight"
-                    attrs="{'invisible':['|', '|', ('qty_to_deliver', '=', 0), ('state', '!=', 'approved'), ('delivery_policy', '=', 'no')]}"
+                    attrs="{'invisible':['|', '|', '|', ('qty_to_deliver', '=', 0), ('qty_to_deliver', '&lt;', 0), ('state', '!=', 'approved'), ('delivery_policy', '=', 'no')]}"
```
florian-dacosta, line 124: Why not ('qty_to_receive', '&lt;=', 0) instead of adding a tuple in  the domain for the same field ?

rma/wizards/rma_make_picking_view.xml
```diff
@@ -108,26 +108,26 @@
                     name="%(action_rma_picking_in)d"
                     string="Create Incoming Shipment"
                     class="oe_highlight"
-                    attrs="{'invisible':['|', '|', ('qty_to_receive', '=', 0), ('state', '!=', 'approved'), ('receipt_policy', '=', 'no')]}"
+                    attrs="{'invisible':['|', '|', '|', ('qty_to_receive', '=', 0), ('qty_to_receive', '&lt;', 0), ('state', '!=', 'approved'), ('receipt_policy', '=', 'no')]}"
```
florian-dacosta, line 111: Why not ('qty_to_receive', '&lt;=', 0) instead of adding a tuple in  the domain for the same field ?

florian-dacosta (commented):
Minor comments, else LGTM

rma/models/rma_order_line.py
```diff
@@ -98,12 +101,24 @@ def _get_rma_move_qty(self, states, direction="in"):
...
-            for move in rec.move_ids.filtered(
-                lambda m: m.state in states and op(m.location_id.usage, rec.type)
-            ):
+                moves = rec.move_ids.filtered(
+                    lambda m: m.state in states
+                    and (
+                        m.location_id.usage == "supplier"
+                        or m.location_id.usage == "customer"
+                    )
+                    and m.location_dest_id.usage == "internal"
```
OriolMForgeFlow:
```suggestion
                    and (m.location_dest_id.usage == "internal"
                        or m.location_dest_id.usage == "supplier"
                    )
```
When the RMA is customer to supplier it is not getting the moves correctly, because the location_id.usage is "customer" but the location_dest_id.usage is "supplier". and does not take into account this option.

OriolMForgeFlow: Also change the following function, it seems that the field "supplier_rma_line_ids" never gets filled.
```
@api.model
     def _get_supplier_rma_qty(self):
-        return sum(
-            self.supplier_rma_line_ids.filtered(lambda r: r.state != "cancel").mapped(
-                "product_qty"
-            )
-        )
+        return self._get_rma_move_qty("done", direction="in")
```

rma/models/rma_order_line.py
```diff
@@ -98,12 +101,24 @@ def _get_rma_move_qty(self, states, direction="in"):
...
+                )
+            elif direction == "out":
+                moves = rec.move_ids.filtered(
+                    lambda m: m.state in states
+                    and (
+                        m.location_dest_id.usage == "supplier"
+                        or m.location_dest_id.usage == "customer"
+                    )
+                    and m.location_id.usage == "internal"
```
OriolMForgeFlow:
```suggestion
                    and (m.location_id.usage == "internal"
                       or m.location_id.usage == "supplier"
                    )
```

When the RMA is 'Dropship deliver to customer' it is not getting the moves correctly, because the location_dest_id.usage is "customer" but the location_id.usage is "supplier" , and does not take into account this option.

DavidJForgeFlow:
@florian-dacosta @OriolMForgeFlow all Done! thank you both!

PR #426, opened 2023-04-03
[14.0][FIX] rma: outgoing moves and pickings

DavidJForgeFlow:
When computing the outgoing moves, it were taking all the pickings that weren't going to internal. An internal move from WH/Stock to VirtualLocations/Scrap will be taked with this method. Now it takes all the moves going from an internal location to customer or supplier. (Also the dropshipping moves).

@ForgeFlow


AaronHForgeFlow (approved):

AaronHForgeFlow:
@LoisRForgeFlow is this fine to you?

LoisRForgeFlow (approved):
LGTM

PR #427, opened 2023-04-04
[16.0][FIX] rma_req_approval_permission

SergiCForgeFlow:
Fixing the fact that any internal user could approve a Customer/Supplier RMA. With this change, only the users who have RMA Manager access rights can perform this action.


AaronHForgeFlow (commented):
I would prefer to change the groups in the view

SergiCForgeFlow:
> I would prefer to change the groups in the view
Yes! Just applied this change.

AaronHForgeFlow (approved):

PR #429, opened 2023-04-17
[16.0][FIX] rma : Do not cancel whole in/out picking when cancelling a rma line

florian-dacosta:
Hello,

There is an important problem with rma line  cancelation.
If you create a rma group, with multiple lines, then create a delivery for all the lines of the group, and then, cancel one line, the whole delivery is canceled. So the rest of the valid (not  canceled) lines won't be shipped (nor canceled)

This PR fixes this, I am not sure why it was done this way, so, reviews are welcome.

An additional problem IMO, is that in case of delivery with multiple steps (with pull rules involved), canceling the rma line only cancel the delivery and the previous steps remains pending.
Any idea to solve this easily ?
@DavidJForgeFlow @LoisRForgeFlow @JordiBForgeFlow

### DIFF
--- a/rma/models/rma_order_line.py
+++ b/rma/models/rma_order_line.py
@@ -666,10 +666,6 @@ class RmaOrderLine(models.Model):
             order.check_cancel()
             order.write({"state": "canceled"})
             order.move_ids._action_cancel()
-            shipments = order._get_in_pickings()
-            shipments |= order._get_out_pickings()
-            for ship in shipments:
-                ship.action_cancel()
         return True
 
     def _get_price_unit(self):


DavidJForgeFlow:
Hi @florian-dacosta, the solution you bring I don't think is correctly at all, because the cancelation of all shipments was also deleting the multiple steps pickings. We also still need to cancel the pickings related to the cancelled RMA. So trully we arrived to a problem. If something occurs to me I would let you know.

florian-dacosta:
> because the cancelation of all shipments was also deleting the multiple steps pickings.

That is not what I saw. I have a case with multiple step pickings and only the delivery is canceled with the current code.

> We also still need to cancel the pickings related to the cancelled RMA

Cancelling the stock move will automatically cancel the picking if there is not other uncanceled moves inside

If the goal is really to cancel everything in case of multiple step pickings, we could do it by checking and canceling the move_orig_ids recursively.

florian-dacosta:
Hello,
This is an important bug around cancelation IMHO, would you mind make a review on this one @LoisRForgeFlow @AaronHForgeFlow ?

AaronHForgeFlow:
We have the need of propagating the cancel when multiple steps pickings in RMA. Otherwise the other pickings remain.

Perhaps we can use a flag somewhere so this can be configured.

florian-dacosta:
> We have the need of propagating the cancel when multiple steps pickings in RMA. Otherwise the other pickings remain.
> 
> Perhaps we can use a flag somewhere so this can be configured.

I have the same need. It currently have a small custom to manage this.
What I don't get, is that I don't see how the current code take care of canceling other picking in case of multiple steps.
And when I test, it does not cancel the other pickings in a multiple step picking process.

I really don't understand the goal of the code I want to remove it since it does only cancel the whole pickings of the `rma.order.line.move_ids`but not the destination or orig pickings. 
For me it really does not make sense when you are using the `rma.group` because if you cancel a line, you don't expect canceling the whole group!

Anyway, I'll try another proposition in an other PR then

PR #431, opened 2023-04-26
[14.0][IMP] rma_scrap: now works with lots/serials

DavidJForgeFlow:
Now when the RMA order line has lots / serial numbers, the scrap flow will use them as default.

@ForgeFlow


LoisRForgeFlow:
@SergiCForgeFlow Can you review this one and if it is fine, start the fwport to v15 and v16?

AaronHForgeFlow (approved):

PR #432, opened 2023-04-28
[14.0][IMP] rma: add date to rma_order_line

DavidJForgeFlow:
Adds date to rma lines. Missing migration script to get default values in already existing databases.


AaronHForgeFlow (commented):
When the date_rma is set in the rma group all the lines should take that date I think.

AaronHForgeFlow:
@DavidJForgeFlow  as long the date is in the rma group I think it makes sense to be consistent with the date in the lines, do you agree?

AaronHForgeFlow:
The same PR is merged in v16 so I am merging this here to keep branches even

PR #433, opened 2023-05-16
[14.0][ADD] rma_sale_mrp

ChrisOForgeFlow:
Solve case when return kits, get right price from components assigned on moves related with delivery on sale order line


rma_mrp/__manifest__.py
```diff
@@ -0,0 +1,15 @@
...
+
+{
+    "name": "RMA Manufacturing Integration",
+    "version": "14.0.1.1.0",
+    "license": "LGPL-3",
+    "category": "RMA",
+    "summary": "Integrates RMA with Kit Manufacturing/Sale Process",
+    "author": "ForgeFlow",
+    "website": "https://github.com/ForgeFlow/stock-rma",
+    "depends": ["rma", "rma_sale", "mrp", "sale_mrp"],
```
AaronHForgeFlow: wouldn't be enough rma_sale and sale_mrp?

AaronHForgeFlow (approved):
LGTM, I would call it rma_sale_mrp, but that is a minor thing.

mariadforgeflow (approved):
LGTM!

PR #434, opened 2023-05-25
[14.0][FIX] rma: procurement origin

JoanSForgeFlow:
In the current implementation of Odoo's `_assign_picking() `method in stock.move, there's a conditional check that looks at whether all the moves associated with a picking have the same partner_id and origin. If any move doesn't align with these conditions, the origin of the picking is set to False.

```
        if any(picking.partner_id.id != m.partner_id.id or
                picking.origin != m.origin for m in moves):
            # If a picking is found, we'll append `move` to its move list and thus its
            # `partner_id` and `ref` field will refer to multiple records. In this
            # case, we chose to  wipe them.
            picking.write({
                'partner_id': False,
                'origin': False,
            })
```

In the context of RMA  when we have multiple moves associated with a picking, each coming from a different RMA order line, we encounter a problem. Each move has its origin set as the name of the RMA orde line (`line.name`), so as soon as a second move from a different line is appended to the picking, the origin of the picking is wiped, because it doesn't match the origin of the first move.

In order to prevent the partner_id of the picking from being set to False when there are multiple associated moves, I propose that we change the origin of the procurement from the name of the RMA line to the name of the procurement group (`group.name`). This way, all moves associated with a picking will share the same origin, preserving the origin of the picking and ensuring it doesn't get inadvertently set to False.

@LoisRForgeFlow @AaronHForgeFlow

### DIFF
--- a/rma/tests/test_rma.py
+++ b/rma/tests/test_rma.py
@@ -779,6 +779,8 @@ class TestRma(common.SavepointCase):
         ).create({})
         wizard._create_picking()
         picking = self.rma_supplier_id.rma_line_ids._get_out_pickings()
+        partner = picking.partner_id
+        self.assertTrue(partner, "Partner is not defined or False")
         moves = picking.move_lines
         self.assertEqual(len(moves), 3, "Incorrect number of moves created")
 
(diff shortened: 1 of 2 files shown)


AaronHForgeFlow (approved):
LGTM, I don't think there is any inconvenience with the change

LoisRForgeFlow (commented):
Could you add a test case? to be sure that single RMA keeps line name and multi RMA from group uses group name? Probably you can just add and assert in one of the already existing tests.

rma/tests/test_rma.py
```diff
@@ -779,6 +779,8 @@ def test_04_supplier_rma(self):
         ).create({})
         wizard._create_picking()
         picking = self.rma_supplier_id.rma_line_ids._get_out_pickings()
+        partner = picking.partner_id
+        self.assertNotEqual(partner, False, "Partner is not defined or False")
```
LoisRForgeFlow: An small advice here, to do true/false assert, better user assertTrue/assertFalse.

LoisRForgeFlow (approved):
Other that the small comment I just did, LGTM

PR #439, opened 2023-06-19
[MIG] rma_scrap: Migration to 15.0

SergiCForgeFlow:


LoisRForgeFlow:
Can you include https://github.com/ForgeFlow/stock-rma/pull/442 here?

LoisRForgeFlow:
> Can you include #442 here?

Nevermind, I see you have embeded the change in the migration commit. Ideally, it should have been cherry-picked to keep versions aligned and easy to trace. Merging anyway as v16 is already merged with the same problem, just keep it in mind for next time.

PR #442, opened 2023-06-27
[FIX] rma_scrap: no scrap if qty to scrap = 0.0

SergiCForgeFlow:
After this fix, it is no longer possible to create a scrap with a quantity of 0.0. Additionally, the tests have been accordingly updated.


LoisRForgeFlow (approved):

LoisRForgeFlow:
Change is embedded in migration commit in https://github.com/ForgeFlow/stock-rma/pull/440 and https://github.com/ForgeFlow/stock-rma/pull/439. Thus, nothing to fwport.

PR #444, opened 2023-07-18
[15.0][MIG] rma_analytic

JasminSForgeFlow:


rma_analytic/tests/test_rma_analytic.py
```diff
@@ -0,0 +1,174 @@
...
+                    move.analytic_account_id,
+                    "the analytic account is not propagated",
+                )
+
+    def test_invoice_analytic(self):
+        """Test wizard to create RMA from a customer invoice."""
+        rma_line = (
+            self.env["rma.order.line"]
+            .with_context(customer=True)
+            .new(
```
AaronHForgeFlow: Could you use [the Form class](https://github.com/ForgeFlow/stock-rma/blob/16.0/rma_account/tests/test_rma_account.py#L236) instead of calling the onchange dfirectly?

AaronHForgeFlow (changes requested):
The analytic account should pass to the stock moves, that is not happening.

rma_analytic/tests/test_rma_analytic.py
JasminSForgeFlow: Its update,
Thanks

JasminSForgeFlow:
> The analytic account should pass to the stock moves, that is not happening.

It is fixed,
Thanks

AaronHForgeFlow (approved):
Functional tests

PR #449, opened 2023-08-17
[14.0][IMP] rma: Auto-calculate package for serial products

BernatPForgeFlow:
When returning or delivering a serial product from a RMA, we will calculate by default its package. Later, it can be modified by the user.

### DIFF
--- a/rma/wizards/rma_make_picking.py
+++ b/rma/wizards/rma_make_picking.py
@@ -206,49 +206,51 @@ class RmaMakePicking(models.TransientModel):
         else:
             pickings = self.mapped("item_ids.line_id")._get_in_pickings()
             action = self.item_ids.line_id.action_view_in_shipments()
-        if picking_type == "incoming":
+        # Force the reservation of the RMA specific lot for incoming shipments.
+        # FIXME: still needs fixing, not reserving appropriate serials.
+        for move in pickings.move_lines.filtered(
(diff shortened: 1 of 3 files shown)


MateuGForgeFlow (approved):

rma/wizards/rma_make_picking.py
```diff
@@ -206,49 +206,51 @@ def action_create_picking(self):
         else:
             pickings = self.mapped("item_ids.line_id")._get_in_pickings()
             action = self.item_ids.line_id.action_view_in_shipments()
-        if picking_type == "incoming":
```
AaronHForgeFlow, line 209: why is this the 'if' removed?

BernatPForgeFlow: With this if, you would never reach this point: https://github.com/ForgeFlow/stock-rma/blob/14.0/rma/wizards/rma_make_picking.py#L238

So, the 'for' loop should be for the incoming and outgoing. Then, for the serial products, we need to calculate lot_id, quantities and package regardless of whether it is incoming or outgoing. If not, inconsistencies are created in the stock quantities.

AaronHForgeFlow (approved):
ok thanks LGTM

AaronHForgeFlow (approved):
Functional tests also

AaronHForgeFlow:
@BernatPForgeFlow  could forward port this?

PR #452, opened 2023-09-05
[14.0][IMP] added default operation on rma group, easy setup before rma lines created

ChrisOForgeFlow:
@ForgeFlow


AaronHForgeFlow (approved):
Nice improvement!

PR #456, opened 2023-09-26
[14.0][REF] rma: change rules to route in warehouse

DavidJForgeFlow:
The stock.rule fields are not used anywhere on rma. It's only usability is automatically setting the rules when setting a new warehouse with a RMA system. However, these rules are created by default in the main warehouse by the data. 
I propose to change the rules for routes to keep a good track of the RMA routes in warehouses and avoid duplicates.

@ForgeFlow

### DIFF
--- a/rma/models/rma_operation.py
+++ b/rma/models/rma_operation.py
@@ -12,7 +12,7 @@ class RmaOperation(models.Model):
     def _default_warehouse_id(self):
         company_id = self.env.user.company_id.id
         warehouse = self.env["stock.warehouse"].search(
-            [("company_id", "=", company_id)], limit=1
+            [("company_id", "=", company_id)], order="sequence asc", limit=1
         )
         return warehouse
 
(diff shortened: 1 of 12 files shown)


AaronHForgeFlow (commented):
Overall looks good. I have to do more testing. Just a couple of questions.

If I create a new warehouse and mark the opetion "RMA in this Warehouse" the routes are not created. I have to save and then edit and mark the option in order to get them created.

Then, once created, I could not make them work as they are. I created a customer RMA to be received in the RMA location for the new warehouse, but when I tried to create the incoming shipment the system was not able to fin the route, even when I select the new route in the RMA line

AaronHForgeFlow:
Also, I think the mehtod def _default_routes(self): in rma_operation should be changed so it takes the routes of the default warehouse

AaronHForgeFlow (approved):
Working good now merging soon

PR #463, opened 2023-10-10
[16.0][FIX] rma: correctly display fields in product category view

JordiMForgeFlow:
Before the fix:

image

After the fix:

image


LoisRForgeFlow (approved):
Ty!

PR #466, opened 2023-10-27
[14.0][IMP] Update copier from template

AaronHForgeFlow:
Currently pre-commit is red in 14.0 so merging as soon it is green


AaronHForgeFlow:
Will not be green until copier is updated with latest changes in https://github.com/OCA/oca-addons-repo-template/pull/214

PR #469, opened 2023-10-27
[16.0][FWD][IMP] added default operation on rma group, easy setup before rma lines created

AaronHForgeFlow:
FW port of https://github.com/ForgeFlow/stock-rma/pull/452


AaronHForgeFlow:
Merging to keep branch align. It works good.

PR #472, opened 2023-10-30
[16.0][FWD][REF] rma: change rules to route in warehouse

AaronHForgeFlow:
forward port of https://github.com/ForgeFlow/stock-rma/pull/456

cc @ForgeFlow


AaronHForgeFlow:
Rebased! :rocket:

PR #479, opened 2023-10-31
[16.0][ADD] product_warranty & rma_product_warranty

AaronHForgeFlow:
Compute warranty based on original move date and invoice date

cc @ ForgeFlow


LoisRForgeFlow:
@AaronHForgeFlow was this finally used? worth merging it?

PR #480, opened 2023-11-16
[14.0][FIX] rma: RMA location in the warehouse should not be inside stock

AaronHForgeFlow:
@JordiMForgeFlow


JordiMForgeFlow (approved):
LGTM 👍🏼

PR #481, opened 2023-11-17
[15.0][FWD][FIX] rma: RMA location in the warehouse should not be inside stock

AaronHForgeFlow:
FWD port of https://github.com/ForgeFlow/stock-rma/pull/480/


AaronHForgeFlow:
Merging to keep branches even

PR #492, opened 2024-01-09
[16.0][IMP] rma: Auto-calculate package for serial products

PauBForgeFlow:
Migrated from #449


AaronHForgeFlow:
Tested and it works. Merging to keep branches even.

PR #495, opened 2024-02-26
[16.0][fix] rma: force to use the same lot through rma.operation

JordiBForgeFlow:


JordiBForgeFlow:
@LoisRForgeFlow @AaronHForgeFlow What do you think about this?

AaronHForgeFlow:
ok to me. I would make the fields default=True so it behaves the same as it behaved before the changes.

DavidJForgeFlow:
This changes should be adapted to much the new module dependency stock_restrict_lot. I have done this changes for v17.0, so if we want to propagate this changes we should take into account the other PR (#620)

PR #498, opened 2024-03-05
[14.0][IMP] rma: add rma_line_id to stock.move views

DavidJForgeFlow:
When looking at stock moves created from a RMA group, it may be confusing knowing who's move is for which RMA line. This improvement should avoid these confusions.

@ForgeFlow

### DIFF
--- a/rma/views/stock_view.xml
+++ b/rma/views/stock_view.xml
@@ -13,6 +13,39 @@
         </field>
     </record>
 
+    <record id="view_move_tree" model="ir.ui.view">
+        <field name="name">stock.move.tree - rma_line_id</field>
+        <field name="model">stock.move</field>
+        <field name="inherit_id" ref="stock.view_move_tree" />
+        <field name="arch" type="xml">
+            <field name="company_id" position="after">
+                <field
+                    name="rma_line_id"
+                    optional="hide"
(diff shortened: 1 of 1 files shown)


rma/views/stock_view.xml
```diff
@@ -13,6 +13,39 @@
...
+    <record id="view_move_tree" model="ir.ui.view">
+        <field name="name">stock.move.tree - rma_line_id</field>
+        <field name="model">stock.move</field>
+        <field name="inherit_id" ref="stock.view_move_tree" />
+        <field name="arch" type="xml">
+            <field name="company_id" position="after">
+                <field
+                    name="rma_line_id"
+                    optional="hide"
+                    attrs="{'invisible': [('rma_line_id', '=', False)]}"
```
LoisRForgeFlow, line 25: shouldn't this be "column_invisible"?

DavidJForgeFlow: to add column_invisible, I should add a field in stock.picking that marks it as created from an RMA. I think it adds too much for a simple visual adding. As this line is in optional="hide", I don't think it bothers that much.

LoisRForgeFlow (approved):

LoisRForgeFlow:
@DavidJForgeFlow can you fwport this one too?

PR #493, opened 2024-01-31
[14.0][FIX] rma, add support to handle package used on customer moves

ChrisOForgeFlow:

PR #499, opened 2024-03-08
[16.0][FIX] rma, add support to handle package used on customer moves

PauBForgeFlow:
Migrated from #493

### DIFF
--- a/rma/hooks.py
+++ b/rma/hooks.py
@@ -1,0 +1,78 @@
+# Copyright 2024 ForgeFlow, S.L.
+# License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl.html).
+from odoo.addons.stock.models.stock_picking import Picking
+
+
+# flake8: noqa: C901
+def post_load_hook():
+    def _check_entire_pack(self):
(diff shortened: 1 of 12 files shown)


AaronHForgeFlow:
Rebased. Conflicts avoided but still not working.

rma/hooks.py
```diff
@@ -0,0 +1,79 @@
...
+                                "move_line_ids": [(6, 0, move_lines_to_pack.ids)],
+                                "company_id": picking.company_id.id,
+                            }
+                        )
+                        # Propagate the result package in the next move for disposable packages only.
+                        # Start Hook
+                        if (
+                            pack.package_use == "disposable"
+                            and pack.location_id
+                            != self.env.ref("stock.stock_location_customers")
```
cormaza: should compare with usage field, pack.location_id.usage != 'customer'

LoisRForgeFlow:
For reference, fwport to v17 was already done in https://github.com/ForgeFlow/stock-rma/pull/499 and it was included in v18 migration. No need to port anything if this gets merged.

PR #502, opened 2024-03-12
[17.0][MIG] rma: migration to 17.0

CarlosVForgeFlow:


JordiBForgeFlow:
@AaronHForgeFlow can you review?

AaronHForgeFlow:
@CarlosVForgeFlow can you please include this https://github.com/ForgeFlow/stock-rma/pull/506?

AaronHForgeFlow:
Edit: this is happening in other RMA versions so we can leave is as known issue for now:

When activating rma groups and creating an RMA group, when going to the lines the customer is not passed to the lines:
image

image

Also, when setting a default operation in the RMA group, it is not passed to the lines on the creation.

AaronHForgeFlow:
Functional tests are good so far, tested full flow of receptions and deliveries with different locations involved. Pending to test with more complex route configurations. Also I want to test the the accounting cost of the products returned back to stock.

PR #508, opened 2024-03-25
[16.0][IMP] rma: create incoming shipment from rma group

chafique-delli:
Everything that is created via the wizards at the level of an RMA line, we wish to do at the level of the RMA group, namely:
- create a single receipt order to process several RMA Group lines
- create a single delivery order to process several RMA Group lines
- create a single supplier RMA to process several lines of the RMA Group
- create a single refund to process several lines of the RMA Group
- create a single sale order to process several lines of the RMA Group
I would like to have your opinion on this process.
If this does not pose a problem for you then I will continue my work to extend this to other wizards ('CREATE DELIVERY', 'CREATE SUPPLIER RMA', ....).
@CarlosVForgeFlow , @florian-dacosta , @AaronHForgeFlow


rma/wizards/rma_make_picking_view.xml
```diff
@@ -108,7 +108,7 @@
                     name="%(action_rma_picking_in)d"
                     string="Create Incoming Shipment"
                     class="oe_highlight"
-                    attrs="{'invisible':['|', '|', '|', ('qty_to_receive', '=', 0), ('qty_to_receive', '&lt;', 0), ('state', '!=', 'approved'), ('receipt_policy', '=', 'no')]}"
+                    attrs="{'invisible':['|', '|', ('qty_to_receive', '&lt;', 0), ('state', '!=', 'approved'), ('receipt_policy', '=', 'no')]}"
                     type="action"
```
florian-dacosta: Why this change ? Reading the code, it seems that the create incoming shipment button will be displayed twice if qty_to_receive = 0

rma/wizards/rma_make_picking_view.xml
```diff
@@ -134,4 +134,27 @@
...
+                    name="%(action_rma_picking_in)d"
+                    string="Create Incoming Shipment"
+                    class="oe_highlight"
+                    attrs="{'invisible':['|', '|', ('qty_to_receive', '=', 0), ('qty_to_receive', '&lt;', 0), ('state', '!=', 'approved')]}"
+                    type="action"
+                />
+                <button
+                    name="%(action_rma_picking_in)d"
+                    string="Create Incoming Shipment"
+                    attrs="{'invisible':['|', '|', ('qty_to_receive', '=', 0), ('qty_to_receive', '>', 0), ('state', '!=', 'approved')]}"
```
florian-dacosta: This button would be hided if qty_to_receive = 0 but displayed if qty_to_receive < 0, it does not seem logic to me.
If it is visible when qty_to_receive < 0, it should also be visible when qty_to_receive = 0.
Also, I believe you need to hide this button if no lines have a receipt policy set, like it is on rma line side.

I believe we have 2 options : 
1) We behave in a different manner than on line side : we display only the receive button when we expected a reception (qty_to_receive > 0) and we don't display the button in the other case. Then, if the user want to force a reception, he should go on rma line where the button is displayed even if we don't expect him to generate a reception.
It may be  a bit confusing for the user to have a different behavior but if seems better/safer to me since on group, we manage multiple lines...

2) Act like on rma line side exactly, then you have to add a field on group with the receipt policy, to hide the button if no policy is set.

@sebastienbeau, it would be nice to have your opinion on this

rma/wizards/rma_make_picking.py
```diff
@@ -34,15 +34,22 @@ def default_get(self, fields_list):
...
+            "rma.order.line",
+            "rma.order",
+        ], "Bad context propagation"
 
         items = []
-        lines = rma_line_obj.browse(rma_line_ids)
+        if active_model == "rma.order":
+            rma = rma_obj.browse(active_ids)
+            lines = rma.rma_line_ids.filtered(lambda x: x.state == "approved")
+        else:
```
florian-dacosta, line 54: I don't think all approved lines should always be added when coming from a group.
I'd say, if the highlighted button is pressed, then we should only get lines having a qty_to_receive > 0
(maybe passing a context on the button for instance ?)
And if the not highlighted button is pressed, then all lines with a receipt policy != 0?

florian-dacosta (changes requested):

rma/wizards/rma_make_picking_view.xml
chafique-delli: @florian-dacosta ,
Reading the calculation method for the 'qty_to_receive' field, it seems to me that qty_to_receive=0.0 if receipt_policy not in ["ordered", "delivered"](so for receipt_policy='no').
This is why I told myself that putting in the domain ('qty_to_receive', '=', 0) and ('receipt_policy', '=', 'no') was a duplicate.

@api.depends(
        "move_ids",
        "move_ids.state",
        "qty_received",
        "receipt_policy",
        "product_qty",
        "type",
    )
    def _compute_qty_to_receive(self):
        for rec in self:
            rec.qty_to_receive = 0.0
            if rec.receipt_policy == "ordered":
                rec.qty_to_receive = rec.product_qty - rec.qty_received
            elif rec.receipt_policy == "delivered":
                rec.qty_to_receive = rec.qty_delivered - rec.qty_received

rma/wizards/rma_make_picking.py
chafique-delli: @florian-dacosta , it's done.

rma/wizards/rma_make_picking_view.xml
florian-dacosta: @chafique-delli But qty_to_receive may be 0  too for receipt_policy in ["ordered", "delivered"]. And in this case, we want to hide this button, as the other not highlighted will be displayed then.

For now with this change, if receipt_policy = "ordered" and qty_to_receive = 0, we have twice the same button (highlighted and not highlighted)

rma/wizards/rma_make_picking_view.xml
sebastienbeau: I think the best solution is the 1. User need a simple UI for processing the 99% of case. So only making visible the right button when there is something to do is the best option.
Then for the advanced case the user can go to the line and make some specific action.

Thanks for your work

chafique-delli:
@AaronHForgeFlow , @JordiBForgeFlow , Please, I would like to have your opinion on processing RMA lines from the RMA group.

AaronHForgeFlow:
@chafique-delli I have not looked much into the code. But the idea of being able to generate a single receipts/deliveries/refunds for a RMA group seems like a good feature to me.

chafique-delli:
> @chafique-delli I have not looked much into the code. But the idea of being able to generate a single receipts/deliveries/refunds for a RMA group seems like a good feature to me.

Ok, if that suits you, I will continue in this direction. THANKS

chafique-delli:
@sebastienbeau , @florian-dacosta , @AaronHForgeFlow , a review please.

AaronHForgeFlow (commented):
Tested so far:

- create a single receipt order to process several RMA Group lines
- create a single delivery order to process several RMA Group lines

It works good. Thanks. Do you plan to continue?

chafique-delli:
> Tested so far:
> 
> * create a single receipt order to process several RMA Group lines
> * create a single delivery order to process several RMA Group lines
> 
> It works good. Thanks. Do you plan to continue?

I have completed the changes planned for the RMA Group.

AaronHForgeFlow:
Thank you @chafique-delli  I will review it soon

AaronHForgeFlow:
@chafique-delli Thank you for the work, only a couple of things:

The refunds for the groups are created good, but for some reason the button in the rma group that takes the user to the created refund it is not working, it does nothing:

image


The sale order creation is correct, and linked correctly to the rma lines of the group. I think a small but nice addition here is to be able to access the generated sales order directly from the group, same as we do for the incoming shipments, deliveries or refunds. Do you think you can do it?

image

Aside from this. I think the only think pending is the discussion about the conditions on the displaying of the buttons.

rma/wizards/rma_make_picking_view.xml
AaronHForgeFlow: I agree with @sebastienbeau in this case. Better to stick to the most common use case.

chafique-delli:
> Aside from this. I think the only think pending is the discussion about the conditions on the displaying of the buttons.

@AaronHForgeFlow , thanks for the review.
For the 'Refunds' button, I'll see why it doesn't work.
For 'Sales' button, I will look to add this to the rma group.

rma_account/wizards/rma_refund.py
```diff
@@ -37,16 +44,24 @@ def default_get(self, fields_list):
...
+        assert active_model in [
+            "rma.order.line",
+            "rma.order",
+        ], "Bad context propagation"
 
         items = []
-        lines = rma_line_obj.browse(rma_line_ids)
+        if active_model == "rma.order":
+            rma = rma_obj.browse(active_ids)
+            lines = rma.rma_line_ids
```
florian-dacosta: shouldn't we have the same logic as for incoming/outgoing shipment creation, take only the line with qty_to_refund > 0?

rma_account/wizards/rma_refund.py
```diff
@@ -90,8 +105,16 @@ def compute_refund(self):
...
+        rma_line_obj = self.env["rma.order.line"]
+        rma_obj = self.env["rma.order"]
+        active_ids = self.env.context.get("active_ids") or []
+        active_model = self.env.context.get("active_model")
+        if active_model == "rma.order":
+            rma = rma_obj.browse(active_ids)
+            lines = rma.rma_line_ids
+        else:
+            lines = rma_line_obj.browse(active_ids)
+        for line in lines:
```
florian-dacosta, line 109: I think this part does not make sense. We check all lines from rma group (or selected rma lines from ui via active_ids) are approved... But then, the refund is created on the base of the wizard item_ids, which may not be related to the checked lines...
Example : select 2 rma lines first one is not approved, second one is ok.
Click on action => create refund. Delete the first line because you changed your mind => click on invoice_refund. It will block because the first line is not approved... But this line won't even be in the refund since it has been deleted from the wizard.

I'd just check the rma lines doing something like `lines = self.item_ids.line_id` instead. It workfs for both case (rma lines/rma group). Does it make sense to you?

rma_sale/wizards/rma_make_sale_order.py
```diff
@@ -166,11 +178,17 @@ class RmaLineMakeSaleOrderItem(models.TransientModel):
...
-            if not self.env.context["active_ids"]:
+            if not active_ids:
                 return
-            rma_line_ids = self.env.context["active_ids"] or []
-            lines = rma_line_obj.browse(rma_line_ids)
+            if active_model == "rma.order":
+                rma = rma_obj.browse(active_ids)
+                lines = rma.rma_line_ids
+            else:
+                lines = rma_line_obj.browse(active_ids)
```
florian-dacosta: I don't understand this whole method. Is this really used ? I think it could be removed fully...
It is a compute method from line_id field to link the item line with the rma_line, but on the loop, it will always take the same rma_line right ? (the first one in active_ids/rma order). So this method assign all item line to the same rma_line.

Then in default_get it is well set...I guess since it is set on default_get, this _compute_line_id is kind of useless, that is why it is working... I'd remove it totally...

rma_sale/wizards/rma_make_sale_order.py
```diff
@@ -43,24 +44,33 @@ def _prepare_item(self, line):
...
+        assert active_model in [
+            "rma.order.line",
+            "rma.order",
+        ], "Bad context propagation"
 
         items = []
-        lines = rma_line_obj.browse(rma_line_ids)
+        if active_model == "rma.order":
+            rma = rma_obj.browse(active_ids)
+            lines = rma.rma_line_ids
```
florian-dacosta: only those we qty_to_sell > 0 ?

rma/wizards/rma_make_picking_view.xml
```diff
@@ -134,4 +134,28 @@
         </field>
     </record>
 
+    <record id="view_rma_button_form" model="ir.ui.view">
```
florian-dacosta: I find it quite confusing to have an override of `rma.order` view in rma_make_picking_view.xml file.
I know there is other similar cases in the rma modules but do we really want to continue this way ? Should it not go to rma_order_view.xml ? @AaronHForgeFlow

florian-dacosta (commented):
@chafique-delli 
I think we have a problem of consistency. There is a different behavior for each button and I find it very confusing.

Incoming shipment and outgoing shipment : Only visible if there is at least one line with a qty to receive/to_deliver
Take by default only rma lines with a qty_to_receive/deliver > 0

Create refund : always visible. Not highlighted if no quantity to refund on any line / highlighted if at least one line with a qty.
Then by default in the wizard, we have all lines, even those which would not have the refund in its operation (in case of different operation by line).

Create SO => Behave like create refund.

In my opinion, we should have the same behavior as for incoming/outgoing shipments. (1 button + take only lines which have a qty to refund / 1 button + only lines with a qty_to_sell > 0)

rma/wizards/rma_make_picking_view.xml
AaronHForgeFlow: @florian-dacosta, better to have file for model, whenever possible, in the next versions some cleanup will be done, hopefully.

chafique-delli:
> @chafique-delli Thank you for the work, only a couple of things:
> 
> The refunds for the groups are created good, but for some reason the button in the rma group that takes the user to the created refund it is not working, it does nothing:
> 
> image
> 
> The sale order creation is correct, and linked correctly to the rma lines of the group. I think a small but nice addition here is to be able to access the generated sales order directly from the group, same as we do for the incoming shipments, deliveries or refunds. Do you think you can do it?
> 
> image
> 
> Aside from this. I think the only think pending is the discussion about the conditions on the displaying of the buttons.

@AaronHForgeFlow , for the refund button, it's fixed and I added the smart button for the sales order.

florian-dacosta (approved):
All seems fine to me now, thanks

sebastienbeau (approved):
LGTM (code review)

florian-dacosta:
@AaronHForgeFlow I believe this PR is ready, when you have time to finish the review!

(I'll port it to 17 when merged)

AaronHForgeFlow:
Nice improvement. Functional tests look good so far. I will ask more people affected by this change to review this.

AaronHForgeFlow (commented):
Create a RMA group with 3 lines, all of them with opreation "Replace after receive". The RMA lines are created find and the Incoming shipment is created fine. However, the button "create Incoming shipment" is still highlighted when if the shipment is already ready to be received:

image

image

If I click on the button again it is proposing me to craete a new incoming shipment for the same quantity, that is not correct:

image

Is it that because of the changes in the "quantity to receive" calculation?

JasminSForgeFlow (approved):
LGTM (code review)

Better to add some tests

florian-dacosta:
> Is it that because of the changes in the "quantity to receive" calculation?

No, I think it is and always has been the behavior, although I agree it is not nice.
If you take the 16.0 branch, create a RMA with operation "Replace after receive", you will have the same issue, until the reception is properly done.
The problem is that the button depend of the field `qty_to_receive` and this `qty_to_receive` value only goes to 0 when the field `qty_received` goes up.
But right after the picking creation, the `qty_received` does not change, but the `qty_incoming` does...

IMO, we should substract the `qty_incoming` AND the quantity `qty_received` from the `qty_to_receive`, it would seems logical to me.
But I don't think it is relative to this PR nor https://github.com/ForgeFlow/stock-rma/pull/553 (got the problem for 1 step reception/expedition)

What do you think ?

florian-dacosta:
Hello @AaronHForgeFlow any news on this one ?
I'd like to port it to 17 then 18, but I'd like it to be merged here first !

AaronHForgeFlow:
Hi @florian-dacosta , I expect to do a last review soon. Last time I checked this it was looking good, I just want to consider the impact of the changes in current production instances.

AaronHForgeFlow:
Hi @florian-dacosta, I am fine with the changes. However, I will do a last check of the impact on the current production instances where these modules are installed. Thank you for you patience. 

Also, a minor thing, I would remove the class oe_highlight from the button "Create Supplier RMA":

image

Separate from this PR, the button "Create sales Quotation" remains highlighted when the quotation is already created, it will be like this until the quotation is confirmed. In my opinion the qty_to_sell field should be counting quotations. But as said, it does not seem part of this PR.

florian-dacosta:
> Hi @florian-dacosta, I am fine with the changes. However, I will do a last check of the impact on the current production instances where these modules are installed. Thank you for you patience.

Thanks you to take the time to review...

> Also, a minor thing, I would remove the class oe_highlight from the button "Create Supplier RMA":

I agree, I've just removed it.

> Separate from this PR, the button "Create sales Quotation" remains highlighted when the quotation is already created, it will be like this until the quotation is confirmed. In my opinion the qty_to_sell field should be counting quotations. But as said, it does not seem part of this PR.

I agree.
Same about the display of the create incoming/outgoing shipment. I think something should be done to avoid to display it when the shipment has been generated already (but has not been validated). 
I'll try to do another PR about this.

florian-dacosta:
Hello @AaronHForgeFlow 

I understand it is a bit complicated to merge that un v16 stable version. 
Do you think that if I propose these changes on v18 existing PRs, we could merge it fast?

The idea would be to leave v16 and v17 alone, but implement this on the not yet merged v18 as it will have less impact on existing instance, what do you think ?

AaronHForgeFlow:
Hi @florian-dacosta I think it is a nice feature and it is ok to me to introduce it in v18. What do you think @DavidJForgeFlow @LoisRForgeFlow ?

LoisRForgeFlow:
As discussed in https://github.com/ForgeFlow/stock-rma/pull/647#issuecomment-3005197961, this improvement reached v18 and  it is a potentially breaking change, so the decision is to not backport it.

PR #510, opened 2024-04-17
[16] rma_sale : Always create rma line from sale line using stock move lines

florian-dacosta:
This way, we have the same way to create the rma lines, wether the tracking is active on the product or not.
Also and mainly, it allows to create rma lines for the component of a KIT sale order line instead of the kit.
When something is sent as a kit, the rma concerns usually only a part of it.

Maybe I could/should make it an option in the module rma_sale_mrp, so one could choose if the add from sale wizard create a line for the kit of for the components. There is a small hook `_create_from_move_line` which can be used for this purpose.

The main consequences of this PR then : 
- In case of kit used in sale order line, the "add from sale" wizard will generate one rma line per components (instead of one rma line for the kit)
- For a normal (not kit) product, with no tracking management, the quantity of the rma line will come from the done stock move line instead of the sale line (same way is is already for the normal (no kit) product with tracking management.

Please, let me know what you think @JordiBForgeFlow @AaronHForgeFlow

### DIFF
--- a/rma_sale/tests/test_rma_sale.py
+++ b/rma_sale/tests/test_rma_sale.py
@@ -65,7 +65,10 @@ class TestRmaSale(common.SingleTransactionCase):
                 "pricelist_id": cls.env.ref("product.list0").id,
             }
         )
-
+        cls.so.action_confirm()
+        for move in cls.so.picking_ids.move_ids:
+            move.write({"quantity_done": move.product_uom_qty})
+        cls.so.picking_ids._action_done()
         # Create RMA group and operation:
         cls.rma_group = cls.rma_obj.create({"partner_id": customer1.id})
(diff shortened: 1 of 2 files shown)


AaronHForgeFlow:
I think both main consequences makes sense, the delivered products are the components of the kit not the kit itself.

AaronHForgeFlow (approved):
Makes sense

AaronHForgeFlow:
Hi @florian-dacosta I checked functionally and I think the change is good. I think that adding the option in the rma_sale_mrp module so user can decide if creating rma lines for the kit components or not is necessary.

florian-dacosta:
Closing, in favor of https://github.com/ForgeFlow/stock-rma/pull/525

PR #514, opened 2024-05-06
[17.0][MIG] rma_scrap

AaronHForgeFlow:
Supersedes https://github.com/ForgeFlow/stock-rma/pull/504

Respected original code & history

Reasons for supersede:

Merge 17.0 was broken after merging and reverting by mistake some pull requests: https://github.com/ForgeFlow/stock-rma/pull/483 & https://github.com/ForgeFlow/stock-rma/pull/500

History was incorrect showing merge commits within it.
Merge conflicts when trying to manually merge other PRs on top on that one.
Unable to see the changes in the migration commit, making hard for the reviewer to review the code

Sorry for the inconvenience.

https://github.com/ForgeFlow


AaronHForgeFlow:
Tested scrap operations with the scrap completion. LGTM

(I am not reviewing myself, the migration was done in the superseded PR)

AaronHForgeFlow:
There is only one issue with the rma_scrap view, with some fields not being displayed (I think they can be removed from the view)

image

RLeeGML:
@AaronHForgeFlow Found a minor issue with creating a scrap order from the RMA order. 

If the product to be scrapped is serial tracked, the scrap form view makes the quantity field readonly. There's an onchange on the product_id field that is supposed to set the quantity to 1 if the product is serial tracked, but it is not triggered. The result is that the quantity is set to 0 and is readonly. Calling the onchange method after creating the scrap order should fix this.

AaronHForgeFlow:
@RLeeOSI thanks for the notice! I will fix that. Anyway, if you create a PR to my branch I would appreciate it.

AaronHForgeFlow:
@RLeeOSI I see that in the Scrap order the quantiy is readonly, but it seems that is standard Odoo. When the product is serial tracked then the quantity is readonly.

In the wizard, I can see I can put a different quantity even if the product is serial:

This quantity is propagated to the scrap order and can't be changed:


So the solution I think is to force the quantity in the wizard to be 1. Same as Odoo does with manual scrap orders.

AaronHForgeFlow:
Rebasing just to confirm tests fail and the module needs an amendment.

rma_scrap/views/rma_order_view.xml
```diff
@@ -0,0 +1,25 @@
...
+                    class="oe_stat_button"
+                    icon="fa-truck"
+                    groups="stock.group_stock_user"
+                    invisible="scrap_count == 0"
+                >
+                    <field name="scrap_count" widget="statinfo" string="Scraps" />
+                </button>
+            </button>
+            <xpath expr="//field[@name='rma_line_ids']/tree" position="inside">
+                <field name="scrap_policy" invisible="True" />
```
DavidJForgeFlow: should be `column_invisible` instead

rma_scrap/models/stock_scrap.py
```diff
@@ -0,0 +1,96 @@
...
+        else:
+            action = self.env.ref("rma.action_rma_supplier_lines")
+            res = self.env.ref("rma.view_rma_line_supplier_form", False)
+        result = action.sudo().read()[0]
+        # choose the view_mode accordingly
+        result["views"] = [(res and res.id or False, "form")]
+        result["res_id"] = self.rma_line_id.id
+        return result
+
+    def action_validate(self):
```
LoisRForgeFlow: @ThiagoMForgeFlow do your comment for v18 also apply here? https://github.com/ForgeFlow/stock-rma/pull/562#pullrequestreview-2756360066

DavidJForgeFlow (changes requested):

rma_scrap/views/rma_order_view.xml
AaronHForgeFlow: done!

rma_scrap/models/stock_scrap.py
AaronHForgeFlow: I removed it, it does not change anything so I think it was a mistake

AaronHForgeFlow: Well, I have to look into it deeper. It seems tests fail if that code is removed (tests does not fail in local :cry: )

LoisRForgeFlow:
@AaronHForgeFlow likely you need this https://github.com/ForgeFlow/stock-rma/pull/623/commits/95665c67e07cf47b8e8e5cff6584c193ac4a470c ??

PR #522, opened 2024-05-22
[16.0][FIX] rma_purchase: add dependency with purchase_stock

### LINKED ISSUE #521: [16.0] Missing dependency for purchase_stock in rma_purchase/__manifest__.py
@max3903 we're using these internally and have just noticed a missing dependency here:

https://github.com/ForgeFlow/stock-rma/blob/16.0/rma_purchase/__manifest__.py#L11

for purchase_stock.

JordiBForgeFlow:
Fixes https://github.com/ForgeFlow/stock-rma/issues/521

### DIFF
--- a/rma_purchase/__manifest__.py
+++ b/rma_purchase/__manifest__.py
@@ -8,7 +8,7 @@
     "license": "LGPL-3",
     "author": "ForgeFlow, Odoo Community Association (OCA)",
     "website": "https://github.com/ForgeFlow/stock-rma",
-    "depends": ["rma_account", "purchase"],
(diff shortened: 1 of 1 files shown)


JordiBForgeFlow:
@AaronHForgeFlow Do you agree with this? We'll need to back and forward port

AaronHForgeFlow (approved):
There is not technical dependency but It makes sense, as the base rma module depends on stock, and we want the purchase_stock features to apply here.

JordiBForgeFlow:
@ArnauCForgeFlow Can you back/forward port?

PR #525, opened 2024-05-23
[16][rma_sale][rma_sale_mrp] Create rma line from stock move line instead of sale order line by default and when possible + manage kit components

florian-dacosta:
I did refactore the way we add rma lines from a sale order.
Today, sometimes quantities are taken from the stock move lines (when tracking management is active for the product) and sometimes quantity is taken from the sale order line (if tracking is not active)
I think it is confusing for the user to have this difference of behavior (which he can't be aware) and, based on the sale order line, we  can't manage the case  where we want to add the components of a kit.

This PR do the following : 
1) Always take the information from the stock move line when possible : the only case where we will take information from the sale order line is when we want to add a kit product.
2) Add an option in rma_sale_mrp, at company level, to choose if we add rma line for the kit product or for the components. (by default, we take the kit, so the present behavior is not  changed)
3) Fix some bugs : 
a) Today, if you have a sale order line with 2 products, but you ship only one and have a backorder (or even cancel the other shipment), the rma line created from the sale order will have a quantity of 2 instead of 1.
b) bug when you use lot and return. Order 1 product with tracking management. Ship it. Then return it, then re-ship it.
When you create the rma line from the sale order, the quantity will be 3, instead of 1.

About rma_sale_mrp, I see it is not set as auto-install. Any reason why ?

@AaronHForgeFlow I did close the previous PR and opened this one because I totally refactored it and added the option on rma_sale_mrp. It seemed cleaner to do a new PR rather than use the other already approved one.

If you could review this one, it would be great !

### DIFF
--- a/rma_sale/wizards/rma_add_sale.py
+++ b/rma_sale/wizards/rma_add_sale.py
@@ -82,7 +82,9 @@ class RmaAddSale(models.TransientModel):
             "target": "new",
         }
 
-    def _prepare_rma_line(self, line, product, quantity, uom_id=False, lot=None):
+    def _prepare_rma_line_from_sale_order_line(
+        self, line, product, quantity, uom_id=False, lot=None
+    ):
         operation = self.rma_id.operation_default_id
         if not operation:
             operation = line.product_id.rma_customer_operation_id
@@ -216,7 +218,7 @@ class RmaAddSale(models.TransientModel):
                         line, existing_sale_line, lot=lot
                     ):
                         continue
-                    data = self._prepare_rma_line(
+                    data = self._prepare_rma_line_from_sale_order_line(
                         line, product, qty, uom_id=uom.id, lot=lot
                     )
                     rec = rma_line_obj.create(data)
@@ -229,7 +231,7 @@ class RmaAddSale(models.TransientModel):
                 if not self._should_create_rma_line(line, existing_sale_line):
                     continue
                 # we can't have lot management based on sale order line
-                data = self._prepare_rma_line(
+                data = self._prepare_rma_line_from_sale_order_line(
                     line,
                     line.product_id,
                     line.product_uom_qty,
--- a/rma_sale/wizards/rma_add_sale.py
+++ b/rma_sale/wizards/rma_add_sale.py
@@ -82,7 +82,7 @@ class RmaAddSale(models.TransientModel):
             "target": "new",
         }
 
-    def _prepare_rma_line_from_sale_order_line(self, line, lot=None):
+    def _prepare_rma_line(self, line, product, quantity, uom_id=False, lot=None):
         operation = self.rma_id.operation_default_id
         if not operation:
             operation = line.product_id.rma_customer_operation_id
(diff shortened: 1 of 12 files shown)


AaronHForgeFlow:
Functional tests are good on my side.  @ChrisOForgeFlow what do you think about this PR?

@florian-dacosta  I think it is ok rma_sale_mrp, if you have installed sale_mrp you are supposed to use kits and compute the cost prices based on the cost of the delivered products. You added a flag in the company so users can decide if using the split by components or not to use it, so I think it is ok to keep it.

AaronHForgeFlow (approved):

rma_sale_mrp/models/res_company.py
```diff
@@ -0,0 +1,9 @@
+# License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl.html)
+
+from odoo import fields, models
+
+
+class ResCompany(models.Model):
+    _inherit = "res.company"
+
+    add_component_from_sale = fields.Boolean()
```
AaronHForgeFlow: Can you rename it to `rma_add_component_from_sale`? so it is clear it is an rma setting

florian-dacosta: Done @AaronHForgeFlow

florian-dacosta:
@AaronHForgeFlow Do you think we can get it merge? Or maybe get another review from forgeflow's team?
When merged I'll gadly port it to v17. It would be nice to have this merged before the 18

AaronHForgeFlow:
Hi @florian-dacosta for me this is fine. The only thing is that we we could split what it is an improvement from what it is a fix it would be great. The method _get_lot_quantity_from_move_lines is for taking the lot for the kit right? It is part of the new feature and not for the fix, right?

florian-dacosta:
@AaronHForgeFlow It would really not be easy to split I think
No `_get_lot_quantity_from_move_lines` is there to compute the quantity we want for the rma line, in the case Odoo looks at the stock.move.line.
This is actually needed for the fix of the 1st bug. But this is also needed in order to be able to add the components of a kit.
So it is needed for both

AaronHForgeFlow:
Ok, I will do a final check with my colleagues and hopefully merge it.

rma_sale/wizards/rma_add_sale.py
```diff
@@ -82,7 +82,7 @@ def select_all(self):
             "target": "new",
         }
 
-    def _prepare_rma_line_from_sale_order_line(self, line, lot=None):
+    def _prepare_rma_line(self, line, product, quantity, uom_id=False, lot=None):
```
ChrisOForgeFlow: Should be keep function name? can be hard to track signature change if original name changes too

ChrisOForgeFlow (changes requested):
LGTM, only one comment, thanks

rma_sale/wizards/rma_add_sale.py
florian-dacosta: Sure, why not
Done

cormaza (approved):
LGTM

ChrisOForgeFlow (approved):
LGTM

AaronHForgeFlow:
Thank you @florian-dacosta , I will keep it mind this MR in this migration https://github.com/ForgeFlow/stock-rma/pull/516, if you create a PR to the branch of that PR I would appreciate it.

florian-dacosta:
Thanks for the help @AaronHForgeFlow 
I'll do the PR in 2 weeks  for version 17 !

PR #528, opened 2024-06-28
[16.0][ADD] rma_product_exchange

chafique-delli:
This module allows you to ship a product different from the one ordered.
@florian-dacosta , @AaronHForgeFlow

### DIFF
--- a/rma_product_exchange/models/rma_order_line.py
+++ b/rma_product_exchange/models/rma_order_line.py
@@ -1,0 +1,30 @@
+# Copyright 2024 Akretion
+# License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl.html)
+from odoo import _, api, fields, models
+from odoo.exceptions import ValidationError
+
+
+class RmaOrderLine(models.Model):
+    _inherit = "rma.order.line"
+
+    new_product_id = fields.Many2one(
+        comodel_name="product.product",
+        tracking=True,
+        help="The new product selected will be shipped "
+        "instead of the product initially ordered.",
+    )
+    product_exchange = fields.Boolean(related="operation_id.product_exchange")
+
+    @api.constrains("new_product_id")
+    def _check_new_product_id(self):
+        for record in self:
+            if (
+                record.new_product_id
+                and record.new_product_id.uom_id != record.product_id.uom_id
+            ):
+                raise ValidationError(
+                    _(
+                        "The selected replacement product must "
+                        "have the same unit of measurement as the initial product !"
+                    )
+                )
--- a/rma_product_exchange/views/rma_operation_view.xml
+++ b/rma_product_exchange/views/rma_operation_view.xml
@@ -1,0 +1,27 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
+
+        <record id="rma_operation_tree" model="ir.ui.view">
+            <field name="model">rma.operation</field>
+            <field name="inherit_id" ref="rma.rma_operation_tree" />
+            <field name="arch" type="xml">
+                <field name="delivery_policy" position="after">
+                    <field name="product_exchange" optional="show" />
+                </field>
+            </field>
+        </record>
+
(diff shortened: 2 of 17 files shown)


AaronHForgeFlow:
For this use case we typically do a sales order free of cost, where we change the product for another. Anyway I think it is interesting to have this use case in the operation and avoid the sales order.

rma_product_exchange/models/procurement_group.py
```diff
@@ -0,0 +1,20 @@
...
+# License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl.html)
+from odoo import api, models
+
+
+class ProcurementGroup(models.Model):
+    _inherit = "procurement.group"
+
+    @api.model
+    def run(self, procurements, raise_user_error=True):
+        if self.env.context.get("rma_item"):
```
florian-dacosta: Is this stuff really usefull ?
You already override `rma_make_picking.wizard._get_product()` it is enough and you don't need to do this stuff with context, or maybe I am missing something

rma_product_exchange/models/rma_order_line.py
```diff
@@ -0,0 +1,13 @@
+# Copyright 2024 Akretion
+# License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl.html)
+from odoo import fields, models
+
+
+class RmaOrderLine(models.Model):
+    _inherit = "rma.order.line"
+
+    new_product_id = fields.Many2one(
```
florian-dacosta, line 10: Maybe a help here to explain what will happen if the field is set ?
Explain that will impact the shipping

rma_product_exchange/views/rma_operation_view.xml
```diff
@@ -0,0 +1,24 @@
...
+                    <field name="product_exchange" optional="show" />
+                </field>
+            </field>
+        </record>
+
+        <record id="rma_operation_form" model="ir.ui.view">
+            <field name="model">rma.operation</field>
+            <field name="inherit_id" ref="rma.rma_operation_form" />
+            <field name="arch" type="xml">
+                <xpath expr="//group[@name='policies']" position="inside">
```
florian-dacosta, line 18: It should be invisible if the delivery_policy is set to "no", because it is only used for delivery, no?

florian-dacosta (commented):
I did not test the module yet. It is hard to see if there won't be any issue having a different product in shipping than in RMA.
For the computation of the qty_received field for instance... 

I think for a start, there should be a constraint somewhere to ensure that the new product has the same uom_id as the base product of the rma line. 
Else I fear we would have quite some bug, starting with this qty computed fields.

rma_product_exchange/models/procurement_group.py
chafique-delli: @florian-dacosta , it's removed.

rma_product_exchange/models/rma_order_line.py
chafique-delli: @florian-dacosta , it's done.

rma_product_exchange/views/rma_operation_view.xml
chafique-delli: @florian-dacosta , it's fixed.

rma_product_exchange/wizards/rma_make_picking.py
```diff
@@ -0,0 +1,36 @@
...
+        if (
+            self.env.context.get("picking_type") == "outgoing"
+            and line.product_exchange
+            and line.new_product_id
+        ):
+            values["product_id"] = line.new_product_id.id
+        return values
+
+    @api.model
+    def _create_procurement(self, item, picking_type):
```
florian-dacosta, line 22: This whole override is useless for now, because `rma_item` context key is not used anywhere anymore, right ?
In if this the case, please, remove it!

florian-dacosta (approved):
One minor comment.
The rest seems ok to me, I'll approve already!

PR #535, opened 2024-08-01
[16.0][ADD] rma_repair_put_away: glue module between rma_repair & rma_put_away

AaronHForgeFlow:
When not using rma_put_away you can define push rules anyway to move the products to the repair location, and those transfers count as rma repair transfers

If you install and use rma_put_away then the rma repair transfers are the ones from the putaway moves

Before this commit there is a dependency issue because the rma_repair code is using a field declared in the rma_put_away, this commit fixes that issue.

### DIFF
--- a/rma_repair/models/rma_order.py
+++ b/rma_repair/models/rma_order.py
@@ -16,7 +16,7 @@ class RmaOrder(models.Model):
         for order in self:
             pickings = (
                 order.mapped("rma_line_ids.move_ids")
-                .filtered(lambda m: m.is_rma_put_away)
+                .filtered(lambda m: m.is_rma_repair_transfer)
                 .mapped("picking_id")
             )
             order.repair_transfer_count = len(pickings)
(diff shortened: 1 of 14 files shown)


JordiMForgeFlow (approved):

MiquelRForgeFlow (approved):

MiquelRForgeFlow:
~~Why not making PR to OCA?~~

AaronHForgeFlow:
@ChrisOForgeFlow do you think this is fine?

ChrisOForgeFlow:
> @ChrisOForgeFlow do you think this is fine?

Yes, thank you

ChrisOForgeFlow (approved):
> @ChrisOForgeFlow do you think this is fine?

Yes, thank you

AaronHForgeFlow:
@ChrisOForgeFlow thank you. I will add a test and merge it.

PR #536, opened 2024-08-12
[17] rma_sale : always take quantity from stock move lines

florian-dacosta:
Port the rma_sale part of : https://github.com/ForgeFlow/stock-rma/pull/525
rma_sale_mrp is not ported nor in PR, so this part will be taken into account when rma_sale_mrp will be migrated.

@AaronHForgeFlow


AaronHForgeFlow:
Thank you @florian-dacosta

PR #539, opened 2024-08-29
[16][IMP] rma : propagate cancelation

florian-dacosta:
Replace https://github.com/ForgeFlow/stock-rma/pull/429

With current code, in the case you create a RMA group with 2 rma lines, if you cancel one rma line, the whole picking related to the rma line is canceled, so the second rma line can't be done.

The goal of this PR is to avoid this, and also to manage the propagation of the cancelation on all origin stock moves related to the rma line. 
For destination moves, it should be configurable on the stock rules.

Like commented in https://github.com/ForgeFlow/stock-rma/pull/429
I don't understand the use case covered with the current code. A example on how to reproduce a process where it is usefull would certainly help me!

I'll set the PR in draft for now, since the goal for now is more to discuss on how can we avoid the problematic of the rma group use while continuing covering the current use case (which I don't get)

@AaronHForgeFlow @DavidFIB 
- If you may try carifying the current use case covered with an example, I would appreciate it.
- If the problematic about the RMA group I explain is not clear for you, I can detail the steps to reproduce the issue.

### DIFF
--- a/rma/models/rma_order_line.py
+++ b/rma/models/rma_order_line.py
@@ -692,12 +692,19 @@ class RmaOrderLine(models.Model):
     def action_rma_cancel(self):
         for order in self:
             order.check_cancel()
+            # cancel ongoing orig moves
+            # dest move cancelation can be managed with propagate_cancel option
+            # on stock rules.
+            moves = order.move_ids
+            to_cancel_orig_moves = self.env["stock.move"]
+            while moves:
+                moves = moves.move_orig_ids.filtered(
+                    lambda m: m.state not in ("done", "cancel") and m.picking_id
+                )
+                to_cancel_orig_moves |= moves
+            to_cancel_orig_moves._action_cancel()
             order.write({"state": "canceled"})
             order.move_ids._action_cancel()
-            shipments = order._get_in_pickings()
-            shipments |= order._get_out_pickings()
-            for ship in shipments:
-                ship.action_cancel()
         return True
 
     def _get_price_unit(self):
--- a/rma/tests/test_rma.py
+++ b/rma/tests/test_rma.py
@@ -1093,3 +1093,94 @@ class TestRma(common.TransactionCase):
         self.assertEqual(
             rma.rma_line_ids.mapped("state"), ["approved", "approved", "approved"]
         )
+
+    def test_10_rma_cancel_line(self):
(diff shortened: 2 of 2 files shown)


DavidJForgeFlow:
Hi @florian-dacosta,
IMO this seems to be a good option, as you say this should fix the behavior of the RMA Groups while maintining the base functionality for the RMA Lines. I would appreciate some tests to make sure the features works as expected, as it wasn't before tested. Thank you.

Also the user @DavidFIB is an outdated profile, so you can tag @DavidJForgeFlow instead!

florian-dacosta:
Thanks for the feedback.
Well, like I said, I did not observe (while testing) the cancelation in chain in case of multi step picking, with the current code.
So I am not sure I "maintain" the base functionality... But the cancelation in chain in case of multi step picking seems to be what is expected so I hope it is fine.

I will add some tests about it, and once ok, will remove the WIP mark of the PR.

florian-dacosta:
@DavidJForgeFlow @AaronHForgeFlow 
Here, I have added a test.
My main goal here is to avoid canceling the whole picking when a single rma line is canceled, in the case of rma.order with multiple lines.

If you play the new test without the changes on `action_rma_cancel` (on present v16), you will see it will fail on a test where the picking is canceled, where it should not. This is what I want to prevent with this PR.

Let me know what you think.

DavidJForgeFlow (approved):
LGTM!

AaronHForgeFlow (approved):
I think this makes sense. Cancelling only the stock moves related to the rma line should be enough. I will do a last check before the merge and hopefully merge it.

florian-dacosta:
Hello @AaronHForgeFlow 
You last check was not conclusive ?

AaronHForgeFlow:
> Hello @AaronHForgeFlow You last check was not conclusive ?

Hi @florian-dacosta sorry for the wait. I think this is ok to merge. However, we have a complex project that may be impacted by this change, and the person handling it is currently on vacation. Hopefully next week we can merge it.

LoisRForgeFlow (approved):
Code review, from my side it makes sense :+1:

AaronHForgeFlow:
@florian-dacosta Thank you for the fix and sorry for the wait, I am merging this now. I did additional functional tests. All good.

PR #540, opened 2024-09-13
[FIX] rma_sale: fix _get_price_unit to handle dropship cases

AlexPForgeFlow:
Take negative valuation layers, as are to customers, and we handle dropship cases where will have negative and positive layers.
Up to now with dropship orders, sums from valuation layers were 0 and div by 0 error was raised.


AaronHForgeFlow (approved):
LGTM

LoisRForgeFlow (approved):
LGTM

PR #550, opened 2024-10-17
[15.0][FIX] rma: stock manager should not inherit rma manager permissions, that group must be explicitly added

AaronHForgeFlow:
cc @ForgeFlow

### DIFF
--- a/rma/security/rma.xml
+++ b/rma/security/rma.xml
@@ -41,10 +41,6 @@
             <field name="category_id" ref="module_category_rma" />
         </record>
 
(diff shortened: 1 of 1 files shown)


DavidJForgeFlow (approved):
IMO makes sense

AaronHForgeFlow:
I think we can merge this too as well. I can do the FWP

DavidJForgeFlow:
@MeritxellAForgeFlow can you FP this to 16.0 and 17.0?

PR #552, opened 2024-11-05
[16.0][FIX] rma: _check_move_partner compare with RMA commercial partner 

AlexPForgeFlow:
When an origin move is selected in an RMA, make the check if the picking address is equal to the delivery address from the RMA customer, instead of the customer address.

### DIFF
--- a/rma/models/rma_order_line.py
+++ b/rma/models/rma_order_line.py
@@ -622,7 +622,8 @@ class RmaOrderLine(models.Model):
         for rec in self:
             if (
                 rec.reference_move_id
-                and rec.reference_move_id.picking_id.partner_id != rec.partner_id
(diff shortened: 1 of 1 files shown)


LoisRForgeFlow (commented):
Makes sense, but maybe it is too restrictive? maybe checking the commercial partner would be enough and do not block cases in which you will deliver to a different location?

AlexPForgeFlow:
> Makes sense, but maybe it is too restrictive? maybe checking the commercial partner would be enough and do not block cases in which you will deliver to a different location?

Yes, you are right. I didn't think about the case of delivering in different addesses.

LoisRForgeFlow (approved):
LGTM!

AaronHForgeFlow (approved):

AaronHForgeFlow:
Perfect! I think we can merge this

PR #553, opened 2024-11-07
[16][FIX] rma: qty received and shipped in case of multiple step reception/expedition

florian-dacosta:
Remove changes introduced here : https://github.com/ForgeFlow/stock-rma/commit/ff0e4fb2147a8069ffd7c0499d1f3cb79808cb5a

Because since this commit, the selection of the move previously in this method have evolved, so now, it breaks the quantity received/shipped in the case of chained moves intead of fixing it...

@AaronHForgeFlow @DavidJForgeFlow

### DIFF
--- a/rma/models/rma_order_line.py
+++ b/rma/models/rma_order_line.py
@@ -125,11 +125,6 @@ class RmaOrderLine(models.Model):
                     )
                 )
             for move in moves:
-                # If the move is part of a chain don't count it
-                if direction == "out" and move.move_orig_ids:
-                    continue
-                elif direction == "in" and move.move_dest_ids:
-                    continue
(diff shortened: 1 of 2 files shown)


DavidJForgeFlow:
Hi @florian-dacosta,
Thanks for the PR! However, can you add some tests to make sure this isn't breaking any feature?

florian-dacosta:
@DavidJForgeFlow 
Sure, I have added a test about qty_received / delivered with multiple steps (chained moves).
The test will fail if you launch on a 16.0 branch. It is fixed in this PR.
Let me know what you think.

DavidJForgeFlow (approved):
Looks good to me! @AaronHForgeFlow can you check this also please!

AaronHForgeFlow (approved):
I think this makes sense. Thank you for the tests.

PR #563, opened 2024-12-02
[16.0][IMP] rma: Calculate expected date in pickings based on the customer lead time

JordiBForgeFlow:
Currently the scheduled date in outgoing pickings is set to today's date, which is unrealistic. We consider a better alternative, which is to add the customer lead time on top of the date when the picking is created.

A more complete setup would probably involve setting up the outgoing lead time in the operation. Then, we can for example estimate 5 days for the customers to send us their product + 5 days to send a replacement = 10 days.

@CarlosVForgeFlow

### DIFF
--- a/rma/wizards/rma_make_picking.py
+++ b/rma/wizards/rma_make_picking.py
@@ -1,11 +1,11 @@
 # Copyright (C) 2017-20 ForgeFlow S.L.
 # License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl.html)
 
-import time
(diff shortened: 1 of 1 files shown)


CarlosVForgeFlow (approved):
LGTM!

DavidJForgeFlow (approved):
Makes sense. Code LGTM and works as well!

DavidJForgeFlow:
@MeritxellAForgeFlow can you please fp to 17.0?

PR #564, opened 2024-12-10
[14.0][FIX] rma: fix reserving not available quantity

DavidJForgeFlow:
Avoid reserving units that are not available. Also ensures the lot picked is the correct one by adding a dependency to stock_move_forced_lot.

Previously, if for example, you try to deliver a unit with a lot not available in you stock, the stock move line will be created and you will no longer able to cancel, delete or validate that picking.

@LoisRForgeFlow

### DIFF
--- a/rma/__manifest__.py
+++ b/rma/__manifest__.py
@@ -10,7 +10,7 @@
     "in odoo",
     "author": "ForgeFlow",
     "website": "https://github.com/ForgeFlow/stock-rma",
-    "depends": ["stock", "mail", "web"],
+    "depends": ["stock", "mail", "web", "stock_move_forced_lot"],
     "demo": ["demo/stock_demo.xml"],
     "data": [
         "security/rma.xml",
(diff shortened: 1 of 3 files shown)


AaronHForgeFlow:
I am not sure about the new dependency. Is it a glue module too hard to be done?

DavidJForgeFlow:
> I am not sure about the new dependency. Is it a glue module too hard to be done?

I think it should be added in the base version, as the way it is working now overrides reservations (leading to create some pickings that cannot be validated nor cancelled or deleted). The changes done by the dependency could be added to RMA but I don't think that makes sense.  
Adding a glue module to fix an error of the base module I think is an error.

AaronHForgeFlow:
Ok. I don't like to add dependencies in stable version but I am afraid there is no "clean" alternative.

LoisRForgeFlow (approved):
LGTM

PR #568, opened 2024-12-16
[17.0] [MIG] rma_reason_code

MeritxellAForgeFlow:
Migration to 17.0
Depends on #529

@ForgeFlow


rma_reason_code/models/reason_code.py
```diff
@@ -0,0 +1,26 @@
...
+class RMAReasonCode(models.Model):
+    _name = "rma.reason.code"
+    _description = "RMA Reason Code"
+
+    def _get_default_color(self):
+        return randint(1, 11)
+
+    name = fields.Char("Code", required=True)
+    description = fields.Text()
+    type = fields.Selection(
```
MiquelRForgeFlow: why not this? 
```suggestion
    something_type = fields.Selection(
```
it's bad trend to use the type field, as already exists [type](https://docs.python.org/3/library/functions.html#type) python function

rma_reason_code/reports/rma_reason_code_report.py
```diff
@@ -0,0 +1,54 @@
...
+
+class RmaReasonCodeReport(models.Model):
+    _name = "rma.reason.code.report"
+    _auto = False
+    _description = "Rma Reason Code Report"
+
+    rma_order_line_id = fields.Many2one(comodel_name="rma.order.line")
+    reason_code_id = fields.Many2one(comodel_name="rma.reason.code")
+    date_rma = fields.Datetime(string="Order Date")
+    type = fields.Selection([("customer", "Customer"), ("supplier", "Supplier")])
```
MiquelRForgeFlow: same here

HviorForgeFlow:
Can remove https://github.com/ForgeFlow/stock-rma/pull/568/commits/735847b8d1b95fe4f5d304dd706cc32097acbb42

Seems code from another module

rma_reason_code/reports/rma_reason_code_report_views.xml
```diff
@@ -0,0 +1,122 @@
...
+              </p><p>
+                Assign a Reason Code to a RTV
+              </p>
+          </field>
+      </record>
+
+      <menuitem
+        action="action_rma_reason_code_report"
+        id="menu_rma_reason_code_report"
+        parent="rma.menu_rma_rma_report"
```
MiquelRForgeFlow, line 111: it's used here @HviorForgeFlow

rma_reason_code/models/reason_code.py
MeritxellAForgeFlow: done!

MiquelRForgeFlow (approved):

HviorForgeFlow (approved):

rma_reason_code/models/reason_code.py
```diff
@@ -13,8 +13,8 @@ def _get_default_color(self):
         return randint(1, 11)
 
     name = fields.Char("Code", required=True)
-    description = fields.Text("Description")
-    type = fields.Selection(
+    description = fields.Text()
+    item_type = fields.Selection(
```
LoisRForgeFlow: now that we do rename the field, I think it is more accurate `rma_type`

MeritxellAForgeFlow: done!

rma_reason_code/models/reason_code.py
```diff
@@ -13,8 +13,8 @@ def _get_default_color(self):
         return randint(1, 11)
 
     name = fields.Char("Code", required=True)
-    description = fields.Text("Description")
-    type = fields.Selection(
+    description = fields.Text()
+    rma_type = fields.Selection(
```
LoisRForgeFlow, line 17: add explicitly the string="RMA Type" in order to use uppercase in RMA.

image

rma_reason_code/views/rma_order_line_views.xml
```diff
@@ -48,7 +48,7 @@
                     widget="many2many_tags"
                     options="{'color_field': 'color', 'no_create': True}"
                 />
-                <field name="allowed_reason_code_ids" invisible="1" />
+                <field name="allowed_reason_code_ids" column_invisible="1" />
```
LoisRForgeFlow: This is a form view, column_invisible only applies to tree/list views

image

LoisRForgeFlow (changes requested):
Some comments below

rma_reason_code/models/reason_code.py
MeritxellAForgeFlow: done!

rma_reason_code/views/rma_order_line_views.xml
MeritxellAForgeFlow: done x2!

LoisRForgeFlow (approved):
LGTM now thanks!

PR #577, opened 2025-02-12
[17.0][MIG] rma: align with v14.0

DavidJForgeFlow:
This PR includes the following changes that do not arrived to v17.0:

- #315 
- #493 
- #564 
- #573 
- #425
- #552 
- #553

PR #579, opened 2025-02-13
[17.0][IMP] rma_filter_lot: show only the lots shipped to the particular customer in the RMA.

DavidJForgeFlow:
Forward port of #287 

Depends on #577


AaronHForgeFlow:
Code LG. Merging to make 17.0 to have all the changes from previous versions.

PR #592, opened 2025-02-20
[17.0][FIX] rma: stock manager should not inherit rma manager permissions, that group must be explicity added

MeritxellAForgeFlow:
Forward port of https://github.com/ForgeFlow/stock-rma/pull/550 to 17.0

@DavidJForgeFlow


DavidJForgeFlow (approved):
LGTM!

PR #596, opened 2025-02-24
[14.0][IMP] rma_sale: hide buttons if count is 0

MeritxellAForgeFlow:
This improvement hides buttons if count is 0.

@DavidJForgeFlow

### DIFF
--- a/rma_sale/views/rma_order_view.xml
+++ b/rma_sale/views/rma_order_view.xml
@@ -12,7 +12,7 @@
                     class="oe_stat_button"
                     icon="fa-pencil-square-o"
                     groups="sales_team.group_sale_salesman"
-                    attrs="{'invisible':[('type', '!=', 'customer')]}"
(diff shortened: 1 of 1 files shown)


rma_sale/views/rma_order_view.xml
```diff
@@ -12,7 +12,7 @@
                     class="oe_stat_button"
                     icon="fa-pencil-square-o"
                     groups="sales_team.group_sale_salesman"
-                    attrs="{'invisible':[('type', '!=', 'customer')]}"
+                    attrs="{'invisible':[('type', '!=', 'customer'), ('sale_count', '=', 0)]}"
```
DavidJForgeFlow: This actually needs an OR condition

DavidJForgeFlow (changes requested):

rma_sale/views/rma_order_view.xml
MeritxellAForgeFlow: That's right, thank you!

DavidJForgeFlow (approved):

PR #599, opened 2025-02-25
[14.0][IMP] rma: more hidden buttons when count is 0

MeritxellAForgeFlow:
This improvement hides buttons when count is 0.

@DavidJForgeFlow


MeritxellAForgeFlow:
The button was already hidden in 14.0, so this can PR can be closed.

PR #603, opened 2025-02-25
[16][IMP] Display product display name on incoming/outgoing stock move instead of rma ref

florian-dacosta:
It  seems more consistent to put product's display name in `stock.move` name rather than a RMA reference because it is what is done in case of sale order : https://github.com/OCA/OCB/blob/16.0/addons/sale_stock/models/sale_order_line.py#L356
Also, we already have the origin field to pass the RMA ref.

I'd like your opinion on this one if possible @AaronHForgeFlow @DavidJForgeFlow

### DIFF
--- a/rma/wizards/rma_make_picking.py
+++ b/rma/wizards/rma_make_picking.py
@@ -125,7 +125,7 @@ class RmaMakePicking(models.TransientModel):
         if not warehouse:
             raise ValidationError(_("No warehouse specified"))
         procurement_data = {
-            "name": line.rma_id and line.rma_id.name or line.name,
+            "name": line.product_id.display_name,
             "group_id": group,
             "origin": group and group.name or line.name,
             "warehouse_id": warehouse,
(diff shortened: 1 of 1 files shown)


AaronHForgeFlow:
It seems legit. However, I am not sure if this can have side effects, like pickings being split by product. Have you checked that?

florian-dacosta:
> It seems legit. However, I am not sure if this can have side effects, like pickings being split by product. Have you checked that?

I am not sure what you mean about pickings split by products, but I am quite sure there are no logic of this kind based on this field `name`

DavidJForgeFlow (approved):
Makes sense and seems it's not grouping procurements by accident. As you say the origin field should be the one containing the RMA ref. So LGTM! Functional review also :+1:

AaronHForgeFlow:
> > It seems legit. However, I am not sure if this can have side effects, like pickings being split by product. Have you checked that?
> 
> I am not sure what you mean about pickings split by products, but I am quite sure there are no logic of this kind based on this field `name`

I thought that could affect the picking creation but I was wrong. So for me we can merge it @DavidJForgeFlow

DavidJForgeFlow:
@florian-dacosta can you please do the FP to 17.0?

PR #617, opened 2025-03-07
[17] Display product display name on incoming/outgoing stock move instead of rma ref

florian-dacosta:
Forward por of https://github.com/ForgeFlow/stock-rma/pull/603

@DavidJForgeFlow


florian-dacosta:
@LoisRForgeFlow This one is merged in v16 + v18 FYI
I don't need it in v17, but it seems better to align

LoisRForgeFlow (approved):
Yes, Thanks!

PR #619, opened 2025-03-12
[16][FIX] Exclude incoming/outgoing quantity from to receive / to deliver quantities

florian-dacosta:
This aims to fix an annoying bug.
Today, the create incoming shipment and delivery button are highlighted when the quantity to receive/to deliver is greater than 0.

The problem is, when you press the button and generate the incoming/outgoing shipment, the picking is generated and the button stay highlighted. This is because the incoming/outcoming quantity does not count when we recompute the to received / to deliver quantity.

It seems logical to me that the to receive / to deliver quantity is set to 0 when the related shipments are generated.
If it is actually wanted / normal to have both to receive and incoming quantity when the incoming shipment is created but not done yet, then I'll change the PR to add 2 computed fields to control the display of the create incoming/outgoing shipments. These computed fields would display the highlighted button if qty_to_receive > 0 AND qty_to_receive > incoming_qy
But my opinion is that it is confusing to have qty to receive and incoming_qty to 1 when you expect 1 product.
It seems like a redondant information.

What do you think @AaronHForgeFlow @DavidJForgeFlow

### DIFF
--- a/rma/models/rma_order_line.py
+++ b/rma/models/rma_order_line.py
@@ -77,7 +77,7 @@ class RmaOrderLine(models.Model):
         for move in self.move_ids:
             first_usage = move._get_first_usage()
             last_usage = move._get_last_usage()
-            if first_usage == "internal" and last_usage != "internal":
+            if first_usage in ("internal", "production") and last_usage != "internal":
                 moves |= move
             elif first_usage == "supplier" and last_usage == "customer":
                 moves |= moves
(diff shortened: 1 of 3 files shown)


AaronHForgeFlow:
I think it makes sense. @DavidJForgeFlow  do you see any potential drawbacks?

florian-dacosta:
Closing in favor of https://github.com/ForgeFlow/stock-rma/pull/656 and https://github.com/ForgeFlow/stock-rma/pull/623

PR #620, opened 2025-03-13
[17.0][IMP] introduce fields in_force_same_lot and out_force_same_lot

DavidJForgeFlow:
Those fields in the rma.operation allows us to control if we want to ensure that the same lot as the one indicated in the RMA should be used in deliveries to customers and receipts from suppliers

Based on #495


LoisRForgeFlow (approved):
Makes sense, and works as expected :+1:

PR #623, opened 2025-03-21
[16.0][IMP] rma: Fix calculation for qty_to_receive

JasminSForgeFlow:
Backport from 18.0

### DIFF
--- a/rma/models/rma_order_line.py
+++ b/rma/models/rma_order_line.py
@@ -132,6 +132,7 @@ class RmaOrderLine(models.Model):
         "move_ids",
         "move_ids.state",
         "qty_received",
+        "qty_incoming",
         "receipt_policy",
         "product_qty",
         "type",
@@ -140,9 +141,13 @@ class RmaOrderLine(models.Model):
         for rec in self:
             rec.qty_to_receive = 0.0
             if rec.receipt_policy == "ordered":
-                rec.qty_to_receive = rec.product_qty - rec.qty_received
+                rec.qty_to_receive = (
+                    rec.product_qty - rec.qty_incoming - rec.qty_received
+                )
             elif rec.receipt_policy == "delivered":
-                rec.qty_to_receive = rec.qty_delivered - rec.qty_received
+                rec.qty_to_receive = (
+                    rec.qty_delivered - rec.qty_incoming - rec.qty_received
+                )
 
     @api.depends(
         "move_ids",
--- a/rma/tests/test_rma.py
+++ b/rma/tests/test_rma.py
@@ -489,8 +489,8 @@ class TestRma(common.TransactionCase):
         )
         # product specific
         self._check_equal_quantity(
-            lines.filtered(lambda l: l.product_id == self.product_1).qty_to_receive,
-            3,
+            lines.filtered(lambda x: x.product_id == self.product_1).qty_to_receive,
+            0,
             "Wrong qty to receive",
         )
         self._check_equal_quantity(
@@ -499,8 +499,8 @@ class TestRma(common.TransactionCase):
             "Wrong qty incoming",
         )
         self._check_equal_quantity(
-            lines.filtered(lambda l: l.product_id == self.product_2).qty_to_receive,
-            5,
+            lines.filtered(lambda x: x.product_id == self.product_2).qty_to_receive,
+            0,
             "Wrong qty to receive",
         )
         self._check_equal_quantity(
@@ -509,8 +509,8 @@ class TestRma(common.TransactionCase):
             "Wrong qty incoming",
         )
         self._check_equal_quantity(
-            lines.filtered(lambda l: l.product_id == self.product_3).qty_to_receive,
-            2,
+            lines.filtered(lambda x: x.product_id == self.product_3).qty_to_receive,
+            0,
             "Wrong qty to receive",
         )
         self._check_equal_quantity(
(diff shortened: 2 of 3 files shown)


AaronHForgeFlow (approved):
For me this is ok, as it is merged in v18. In v16 there is this other proposal: https://github.com/ForgeFlow/stock-rma/pull/619/ but I think we can merge this other because it is already proposed to v17 and merged in v18

JordiMForgeFlow (approved):

LoisRForgeFlow:
Thanks Jasmin! we'll merge after closing v18 migration

florian-dacosta:
Thanks.
Has the same been done for `qty_to_deliver` ?

JasminSForgeFlow:
> Thanks. Has the same been done for `qty_to_deliver` ?

That's working okay, no need for it

florian-dacosta:
> > Thanks. Has the same been done for `qty_to_deliver` ?
> 
> That's working okay, no need for it

Are you sure ? 
As far as I know, the problem is exactly the same.
If you have a rma line with `product_qty = 1`.  the `qty_to_deliver` is also 1 / `qty_outgoing` is 0 Then you generate the delivery order, the `qty_to_deliver` is still 1 and the `qty_outgoing` is also 1 (https://github.com/ForgeFlow/stock-rma/pull/623#issuecomment-2786166328) while the `qty_to_deliver` should probably goes to 0, as it happens with the qty_to_receive in the case of incoming shipments, with this PR, right ?

For me the qty_to_receive and the qty_to_deliver work exactly the same, so the problem is the same, and it would be really weird if the field would some times take into account the incoming/outgoing qty and sometime not !
But I may have misunderstood some part or maybe we do not have the same issue in mind around this field...
Could you elaborate please ?

JasminSForgeFlow:
> > > Thanks. Has the same been done for `qty_to_deliver` ?
> > 
> > 
> > That's working okay, no need for it
> 
> Are you sure ? As far as I know, the problem is exactly the same. If you have a rma line with `product_qty = 1`. the `qty_to_deliver` is also 1 / `qty_outgoing` is 0 Then you generate the delivery order, the `qty_to_deliver` is still 1 and the `qty_outgoing` is also 1 ([#623 (comment)](https://github.com/ForgeFlow/stock-rma/pull/623#issuecomment-2786166328)) while the `qty_to_deliver` should probably goes to 0, as it happens with the qty_to_receive in the case of incoming shipments, with this PR, right ?
> 
> For me the qty_to_receive and the qty_to_deliver work exactly the same, so the problem is the same, and it would be really weird if the field would some times take into account the incoming/outgoing qty and sometime not ! But I may have misunderstood some part or maybe we do not have the same issue in mind around this field... Could you elaborate please ?

Here I have created one RMA order, Qty to receive is 1 and other all are 0

image

I have create shipment, haven't validated yet

image

After validating shipment

image

It show Qty to delivery 1 only after shipment validated.

florian-dacosta:
Hello @JasminSForgeFlow 
Thank you for having a look. But the case you show is not the one I try to solve.
My issue is that the "Qty To Deliver" is not  decreased when the `Outgoing Qty` is increased.
It seems this is the problem you solve about the "Qty to receive" and "Incoming Qty" in this PR, but AFAIK, the exact same problem happens about outgoing quantities.

See the following example, done on v16 + this current PR
Create a RMA for 1 product , as a first step, I just have qty to receive : 1

image

Then, I create the incoming shipment, the qty to receive is now 0, and the incoming qty = 1
=> This is what is fixed by your PR, because before, we would have  qty to receive = 1 and incoming qty = 1.
image

Then I validate the incoming shipment
We are still fine with qty received = 1 and the other incoming qties =0 + qty to receive = 1
image

Then I create the delivery order, and that is where my problems start :+1: 
image

Here, I still have the button to create delivery, but it is already created.
The outgoing quantity is 1 => It is ok
The qty to deliver is still 1 => Not consistent with the way it now works with incoming quantity.

What do you think ?

florian-dacosta:
One last thing, I spot another issue, in case you manufacture product, in case you create a delivery from the RMA but you need to produce it before sending (with some make to order `stock.rule`) the outgoing quantities are not properly computed. I did also fix it in my PR : https://github.com/ForgeFlow/stock-rma/pull/619/files#diff-104cd457814307a572bcee64a3f4ef3a33ef1cab047ea0c00885888c860fc564R80
If we go with yours, it  would be nice to take this part as well.

AaronHForgeFlow:
@florian-dacosta 

> Here, I still have the button to create delivery, but it is already created.
> The outgoing quantity is 1 => It is ok
> The qty to deliver is still 1 => Not consistent with the way it now works with incoming quantity.

On the first design that was kind of "on purpose" because the quantity is not actually delivered. But it is true that the "Create Delivery" button is highlighted and that is not good.

florian-dacosta:
> @florian-dacosta
> 
> > Here, I still have the button to create delivery, but it is already created.
> > The outgoing quantity is 1 => It is ok
> > The qty to deliver is still 1 => Not consistent with the way it now works with incoming quantity.
> 
> On the first design that was kind of "on purpose" because the quantity is not actually delivered. But it is true that the "Create Delivery" button is highlighted and that is not good.

@AaronHForgeFlow 
I understand, I guess one can prefer one or the other design, but in this PR, the behavior is changed for incoming quantity, my point is that it should change for outgoing quantity as well, or it would be totally confusing right ?

rma/models/rma_order_line.py
```diff
@@ -132,6 +132,7 @@ def _get_rma_move_qty(self, states, direction="in"):
         "move_ids",
         "move_ids.state",
         "qty_received",
+        "qty_incoming",
```
florian-dacosta, line 135: Missing qty_delivered in dependencies ?

LoisRForgeFlow (approved):
:+1:

PR #624, opened 2025-04-08
[18][rma] Add possibility to create incoming/delivery orders and customer/supplier rma from rma group

florian-dacosta:
This is a port of https://github.com/ForgeFlow/stock-rma/pull/508 (which is approved already)
But splitted in 3 PR, one per module (`rma`, `rma_account`, `rma_sale`)
The goal is to merge it in this recent version since it may seems a lot of change for a stable version (but it is actually very few changes, things is some refactore in views was also done (move `rma.order `and` rma.order.line` view override from wizard to view folder).

This PR on `rma` add 3 buttons on RMA group : 
- Incoming shipment creation
- Delivery order creation
- customer/supplier rma creation

@AaronHForgeFlow @sebastienbeau @JasminSForgeFlow


DavidJForgeFlow (approved):

AaronHForgeFlow (approved):
:+1:  I think it is a good improvement

PR #626, opened 2025-04-08
[18][rma_sale] Add possibility to create sale order from rma group

florian-dacosta:
This is a port of https://github.com/ForgeFlow/stock-rma/pull/508 (which is approved already)
But splitted in 3 PR, one per module (rma, rma_account, rma_sale)
The goal is to merge it in this recent version since it may seems a lot of change for a stable version (but it is actually very few changes, things is some refactore in views was also done (move rma.order and rma.order.line view override from wizard to view folder).

This PR on rma add 1 button on RMA group :
- Sale Order creation from rma group
- Fix some sale smart button on rma group

@AaronHForgeFlow @sebastienbeau @JasminSForgeFlow

### DIFF
--- a/rma_sale/__manifest__.py
+++ b/rma_sale/__manifest__.py
@@ -13,11 +13,11 @@
     "data": [
         "security/ir.model.access.csv",
         "data/rma_operation.xml",
+        "wizards/rma_make_sale_order_view.xml",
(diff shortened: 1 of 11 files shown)


AaronHForgeFlow (approved):
Functional LGTM!

PR #632, opened 2025-04-23
[17.0][FIX] rma: fixed inconsistencies with delivery quantities

ArnauCForgeFlow:
This pr includes the following fixes:

- Fix the "quantity to deliver" calculation and prevent it from being negative.
- Use "quantity" instead of "demand" to calculate the delivered quantity.

Related issues: https://github.com/ForgeFlow/stock-rma/issues/628 https://github.com/ForgeFlow/stock-rma/issues/629

### DIFF
--- a/rma/models/rma_order_line.py
+++ b/rma/models/rma_order_line.py
@@ -122,7 +122,12 @@ class RmaOrderLine(models.Model):
                     continue
                 elif direction == "in" and move.move_dest_ids:
                     continue
-                qty += product_obj._compute_quantity(move.product_uom_qty, rec.uom_id)
+                if move.state == "done":
+                    qty += product_obj._compute_quantity(move.quantity, rec.uom_id)
+                else:
+                    qty += product_obj._compute_quantity(
+                        move.product_uom_qty, rec.uom_id
+                    )
             return qty
 
     @api.depends(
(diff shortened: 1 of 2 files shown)


rma/models/rma_order_line.py
```diff
@@ -108,7 +108,7 @@ def _compute_out_shipment_count(self):
             pickings = line._get_out_pickings()
             line.out_shipment_count = len(pickings)
 
-    def _get_rma_move_qty(self, states, direction="in"):
+    def _get_rma_move_qty(self, states, direction="in", use_demand_qty=True):
```
LoisRForgeFlow: better than adding an argument to the method, do it conditionally based on the state, if the stock move is done use `quantity` if not use `product_uom_qty`

rma/models/rma_order_line.py
```diff
@@ -122,7 +122,12 @@ def _get_rma_move_qty(self, states, direction="in"):
                     continue
                 elif direction == "in" and move.move_dest_ids:
                     continue
-                qty += product_obj._compute_quantity(move.product_uom_qty, rec.uom_id)
+                if use_demand_qty:
+                    qty += product_obj._compute_quantity(
+                        move.product_uom_qty, rec.uom_id
+                    )
+                else:
+                    qty += product_obj._compute_quantity(move.quantity, rec.uom_id)
```
LoisRForgeFlow: here, decide based on status

LoisRForgeFlow (changes requested):

rma/models/rma_order_line.py
ArnauCForgeFlow: Makes sense! Done.

LoisRForgeFlow: EDITED: nevermind, my mistake.

LoisRForgeFlow (changes requested):
@ArnauCForgeFlow I just rebased, I tested and there is still a corner case possible:

I have this delivery for more units than actually needed (4 instead of 1), also 1 unit is reserved.

image

The RMA still show 1 unit pending to deliver, instead of 4 outgoing.

image

ArnauCForgeFlow:
@LoisRForgeFlow Could you check if now its working as you expected?

Thanks!!

LoisRForgeFlow:
Rebased.

LoisRForgeFlow (approved):
Tested, it looks good now :+1:

LoisRForgeFlow:
@ArnauCForgeFlow could you port to 16 and 18? Thanks!

PR #573, opened 2025-01-21
[14.0][REF] rma: change stock_move_forced_lot to stock_restrict_lot

DavidJForgeFlow:
The two modules have the same features, but the stock_restrict_lot is more used in other modules so changing the module used to match the OCA behavior.

PR #634, opened 2025-04-24
[16.0][FWD][IMP] rma: align with v14.0

ArnauCForgeFlow:
This PR includes the following changes that do not arrived to v16.0:
- #315
- #493
- #564
- #573
- #425

### DIFF
--- a/rma/__manifest__.py
+++ b/rma/__manifest__.py
@@ -3,13 +3,13 @@
 
 {
     "name": "RMA (Return Merchandise Authorization)",
-    "version": "16.0.1.0.0",
+    "version": "16.0.1.1.0",
     "license": "LGPL-3",
     "category": "RMA",
     "summary": "Introduces the return merchandise authorization (RMA) process in odoo",
     "author": "ForgeFlow",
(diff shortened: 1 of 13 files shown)


DavidJForgeFlow (approved):
Functional review :+1:

LoisRForgeFlow:
For reference, based on https://github.com/ForgeFlow/stock-rma/pull/577

rma/__manifest__.py
```diff
@@ -3,13 +3,13 @@
 
 {
     "name": "RMA (Return Merchandise Authorization)",
-    "version": "16.0.1.0.0",
```
LoisRForgeFlow, line 6: @ArnauCForgeFlow a bit nitpicking but could you bump at least the minor? (16.0.1.1.0 - also update the script folder). We are adding a new module dependency and changing the data model, it is for sure not a patch.

ArnauCForgeFlow: You are right, done!

LoisRForgeFlow (approved):
Thanks :+1:

PR #635, opened 2025-04-24
 [16.0][BP][IMP] introduce fields in_force_same_lot and out_force_same_lot

ArnauCForgeFlow:
Backport of #620 

Depends on: #634

### DIFF
--- a/rma/models/stock_move.py
+++ b/rma/models/stock_move.py
@@ -62,6 +62,7 @@ class StockMove(models.Model):
             not lot_id
             and self.rma_line_id.lot_id
             and self.location_id.usage == "internal"
+            and self.rma_line_id.operation_id.out_force_same_lot
         ):
             # In supplier RMA deliveries we can only send the RMA lot/serial.
             lot_id = self.rma_line_id.lot_id
@@ -88,6 +89,7 @@ class StockMove(models.Model):
             not lot_id
             and self.rma_line_id.lot_id
             and self.location_id.usage == "internal"
+            and self.rma_line_id.operation_id.out_force_same_lot
         ):
             # In supplier RMA deliveries we can only send the RMA lot/serial.
             lot_id = self.rma_line_id.lot_id
--- a/rma/__manifest__.py
+++ b/rma/__manifest__.py
@@ -3,7 +3,7 @@
 
 {
     "name": "RMA (Return Merchandise Authorization)",
-    "version": "16.0.1.1.0",
+    "version": "16.0.1.2.0",
     "license": "LGPL-3",
     "category": "RMA",
     "summary": "Introduces the return merchandise authorization (RMA) process in odoo",
--- a/rma/data/rma_operation.xml
+++ b/rma/data/rma_operation.xml
@@ -8,6 +8,8 @@
         <field name="type">customer</field>
         <field name="in_route_id" ref="rma.route_rma_customer" />
         <field name="out_route_id" ref="rma.route_rma_customer" />
+        <field name="in_force_same_lot">True</field>
+        <field name="out_force_same_lot">False</field>
     </record>
 
     <record id="rma_operation_supplier_replace" model="rma.operation">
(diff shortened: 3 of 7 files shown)


DavidJForgeFlow:
Functional review :+1: Take care as you are also adding the same commits as #634 !

LoisRForgeFlow:
@ArnauCForgeFlow you can rebase now that #634 is merged

ArnauCForgeFlow:
@LoisRForgeFlow done!!

LoisRForgeFlow (approved):
Thanks :+1:

rma/models/stock_move.py
```diff
@@ -62,6 +62,7 @@ def _get_available_quantity(
             not lot_id
             and self.rma_line_id.lot_id
             and self.location_id.usage == "internal"
+            and self.rma_line_id.operation_id.out_force_same_lot
```
florian-dacosta, line 65: I do see this PR a bit late but, are you sure this part is still necessary ?
Indeed, I see this PR now fill the `restrict_lot_id` in the procurement data, and then the module `stock_restrict_lot` should manage the reservations properly.

rma/models/stock_move.py
```diff
@@ -88,6 +89,7 @@ def _update_reserved_quantity(
...
```
florian-dacosta, line 92: I do see this PR a bit late but, are you sure this part is still necessary ?
Indeed, I see this PR now fill the `restrict_lot_id` in the procurement data, and then the module `stock_restrict_lot` should manage the reservations properly.

florian-dacosta (commented):
@ArnauCForgeFlow
Could you check my comments please ?
Also, in line with my PR : https://github.com/ForgeFlow/stock-rma/pull/615
I believe that now that we add the `restrict_lot_id`, we could remove the whole logic here : https://github.com/ForgeFlow/stock-rma/blob/16.0/rma/wizards/rma_make_picking.py#L235
Right ?

DavidJForgeFlow:
> @ArnauCForgeFlow Could you check my comments please ? Also, in line with my PR : #615 I believe that now that we add the `restrict_lot_id`, we could remove the whole logic here : https://github.com/ForgeFlow/stock-rma/blob/16.0/rma/wizards/rma_make_picking.py#L235 Right ?

That part can be modified yes, but not completely removed as it also contains some logic for the packages.
But yes, I think this can be edited to let the reservations of SN work only by the restrict lot.
@ArnauCForgeFlow will check it ASAP.

rma/models/stock_move.py
DavidJForgeFlow: In this case, what this is doing is being able to configure if the restrict lot is enabled or not by rma.operation and incoming or outgoing pickings. So this specific line is needed.

florian-dacosta: Depending on the `out_force_same_lot` option of the `rma.operation`, the `restrict_lot_id` is or is not set on the `stock.move` : https://github.com/ForgeFlow/stock-rma/blob/16.0/rma/wizards/rma_make_picking.py#L142

Then, AFAIK, the `stock_restrict_lot` module take care of the reservation. Here, I feel like we do it twice...

DavidJForgeFlow: You are right, my mistake here. @ArnauCForgeFlow can you please delete L65 and L92 in the new PR please!

florian-dacosta: Shouldn't we delete the whole override ?
Because for me the problem is the lines 68 and 95. We do not need (I think) to force a lot in this module because it should be done already by `stock_restrict_lot`
If you juste delete L65 and L92, it is worse because it will force the RMA lot no matter if it is set on the `restrict_lot_id`.
See https://github.com/OCA/stock-logistics-workflow/blob/16.0/stock_restrict_lot/models/stock_move.py#L46

DavidJForgeFlow: Yes you are right, the two methods should be removed. @ArnauCForgeFlow

PR #636, opened 2025-04-25
[17.0][FIX] rma: fixed warehouse rule and route creation

ArnauCForgeFlow:
- This PR inherits the create method to apply the same logic as in write when generating warehouse stock.rule records. 
- It also removes demo data to prevent duplicate rule creation.
- Moves the field rma_in_this_wh to a new group called "RMA" because resupply group is only visible when having more than 1 warehouse.

Related issue: https://github.com/ForgeFlow/stock-rma/issues/627


rma/models/stock_warehouse.py
```diff
@@ -93,6 +93,32 @@ def write(self, vals):
...
+                    )
+                # RMA types
+                if not wh._rma_types_available():
+                    wh._create_rma_picking_types()
+                else:
+                    for r_type in wh._get_rma_types():
+                        if r_type:
+                            r_type.active = True
+                # RMA rules:
+                wh._create_or_update_rma_pull()
```
LoisRForgeFlow: @ArnauCForgeFlow Please, do not duplicate code, move the changes to a method that you can call from both `create` and `write`

LoisRForgeFlow (changes requested):

rma/models/stock_warehouse.py
ArnauCForgeFlow: @LoisRForgeFlow You are right, done!

rma/tests/test_rma.py
```diff
@@ -1259,7 +1251,7 @@ def test_10_rma_cancel_line(self):
         wizard._create_picking()
         # cancel first line, check both chained move are canceled
         second_rma_out_move = second_rma_line.move_ids.filtered(
-            lambda m: m.picking_id.picking_type_code == "outgoing"
```
LoisRForgeFlow: I don't understand this change.

Just in case I checked in v16, to confirm they are not equivalent:

v16


v17

image

LoisRForgeFlow (changes requested):

rma/tests/test_rma.py
ArnauCForgeFlow: Yes, that doesn't make sense, fixed @LoisRForgeFlow

LoisRForgeFlow (approved):
OK, this looks good now!

LoisRForgeFlow:
@ArnauCForgeFlow could you fwport to 18 and backport to 16?

PR #637, opened 2025-05-05
[IMP] Add hook for forcing reservation in incoming shipments

florian-dacosta:
This is a trivial PR  where I only put the reservation logic on incoming shipment in a separate method.
The goal is to be able to merge this (and port this) quite fast/easily since the refactore/fix of this part does not seem so simple.

I had start a PR https://github.com/ForgeFlow/stock-rma/pull/615, which is obsolete now because of this merge : https://github.com/ForgeFlow/stock-rma/pull/634
But this logic still seems quite buggy for me and I need a way to be able to bypass it.
It seems legit to have this in a separated method anyway.

Do you think we can go forward fast with it ?
The diff is big but I actually did not make any change in the code on this one, just moved it.
@LoisRForgeFlow @AaronHForgeFlow @DavidJForgeFlow

### DIFF
--- a/rma/wizards/rma_make_picking.py
+++ b/rma/wizards/rma_make_picking.py
@@ -223,104 +223,105 @@ class RmaMakePicking(models.TransientModel):
             return True
         return False
 
-    def action_create_picking(self):
-        self._create_picking()
+    def _force_reservation_specific_lot(self, pickings, picking_type):
         move_line_model = self.env["stock.move.line"]
-        picking_type = self.env.context.get("picking_type")
-        if picking_type == "outgoing":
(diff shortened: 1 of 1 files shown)


DavidJForgeFlow:
It looks like a good idea, once #639 is finished and merged, you could do the same change or write into that PR to make @ArnauCForgeFlow do the change instead. Thanks @florian-dacosta

DavidJForgeFlow:
The commit is added in #639! if you think it looks good, maybe it's interesting to close this one and merge everything in the other one! @florian-dacosta

florian-dacosta:
Sure, I close this one since the commit has been added in the other, thanks

PR #656, opened 2025-06-24
[16.0][rma] Fix dependency of qty_to_deliver field + take into account production usage as internal to compute outgoing moves

florian-dacosta:
Replace https://github.com/ForgeFlow/stock-rma/pull/619 now that https://github.com/ForgeFlow/stock-rma/pull/654 has been merged

@LoisRForgeFlow @AaronHForgeFlow @ArnauCForgeFlow

### DIFF
--- a/rma/models/rma_order_line.py
+++ b/rma/models/rma_order_line.py
@@ -77,7 +77,7 @@ class RmaOrderLine(models.Model):
         for move in self.move_ids:
             first_usage = move._get_first_usage()
             last_usage = move._get_last_usage()
-            if first_usage == "internal" and last_usage != "internal":
+            if first_usage in ("internal", "production") and last_usage != "internal":
                 moves |= move
             elif first_usage == "supplier" and last_usage == "customer":
                 moves |= moves
@@ -156,6 +156,7 @@ class RmaOrderLine(models.Model):
         "type",
         "qty_delivered",
         "qty_received",
+        "qty_outgoing",
     )
     def _compute_qty_to_deliver(self):
         for rec in self:


rma/models/rma_order_line.py
```diff
@@ -77,7 +77,7 @@ def _get_out_moves(self):
         for move in self.move_ids:
             first_usage = move._get_first_usage()
             last_usage = move._get_last_usage()
-            if first_usage == "internal" and last_usage != "internal":
+            if first_usage in ("internal", "production") and last_usage != "internal":
```
LoisRForgeFlow, line 80: @florian-dacosta could you remind me what was the case that motivated this fix?

florian-dacosta: Case of manufacturer on demand (make to order)
I have a RMA where I want to send a new product to my customer, I'll have a delivery order and a manufacture order (thanks to the route/pull rule MTO configuration). I could have any more step between the delivery and the manufacturing of course.

So when this `_get_out_moves` method is called, which is used to compute the delivered and outgoing quantity, without this fix, Odoo won't find any "out move" because it will find "production" on first usage and "!= internal" as last usage. And this is a none existing case in this method.

So with my case, without the fix, when I use the button "create delivery", the `Outgoing Qty`, `Qty To Deliver` `Qty Delivered` are not updated.
And when I produce the MO and validate the delivery order, same, no change on these outgoing qty fields.

So that fixes this use case. It seems quite logical that production is considered as an internal usage.

Is it clearer ? @LoisRForgeFlow

LoisRForgeFlow: Yep! thanks for the clear explanation :+1:

LoisRForgeFlow (approved):

PR #590, opened 2025-02-20
[17.0][FIX] rma*: change invisible for column_invisible in tree views when trying to hide the full column

DavidJForgeFlow:
@AaronHForgeFlow @LoisRForgeFlow 

Pre-fix:
image
Post-fix:
image

PR #657, opened 2025-06-25
[18.0][FWP][FIX] rma*: change invisible for column_invisible in tree views when trying to hide the full column

ArnauCForgeFlow:
Forward port of #590 

- rma_repair commit not needed.
- rma_sale commit not needed.


LoisRForgeFlow (approved):
Thanks!

PR #659, opened 2025-06-25
[17][FWP][FIX] dependency of qty_to_deliver field + take into account production usage as internal to compute outgoing moves

florian-dacosta:
forward port of https://github.com/ForgeFlow/stock-rma/pull/656

PR for 18 following right after @LoisRForgeFlow


LoisRForgeFlow (approved):
Ty!

PR #660, opened 2025-06-25
[18][FWP][FIX] dependency of qty_to_deliver field + take into account production usage as internal to compute outgoing moves

florian-dacosta:
Forward port of https://github.com/ForgeFlow/stock-rma/pull/656
@LoisRForgeFlow


LoisRForgeFlow (approved):
Thanks!

PR #661, opened 2025-06-25
[17.0][FWP][rma] Fix dependency of qty_to_deliver field + take into account production usage as internal to compute outgoing moves

ArnauCForgeFlow:
Forward port of #656


ArnauCForgeFlow:
Closing as Florian already done the ports

PR #662, opened 2025-06-26
[18.0][IMP] Display product display name on incoming/outgoing stock move instead of rma ref

LoisRForgeFlow:
Fwport of #603 

@florian-dacosta it was actually missing in v18 too.


florian-dacosta (approved):
Hum, I checked, but I must have done a mistake because of the 16.0 default branch maybe, my bad !

PR #663, opened 2025-06-30
[16.0][IMP] rma_purchase: allow to specify picking type

AaronHForgeFlow:
For example, I want to purchase on a specific warehouse or location. My use case is subcontractors are sending the products from specific location depending on the purpose of the purchase

image

Is it ok to put this in the base module? or it may confuse users?

### DIFF
--- a/rma_purchase/wizards/rma_order_line_make_purchase_order.py
+++ b/rma_purchase/wizards/rma_order_line_make_purchase_order.py
@@ -16,6 +16,11 @@ class RmaLineMakePurchaseOrder(models.TransientModel):
         required=False,
         readonly=1,
     )
+    picking_type_id = fields.Many2one(
+        comodel_name="stock.picking.type",
+        string="Picking Type",
+        default=lambda self: self._default_picking_type(),
+    )
     item_ids = fields.One2many(
         comodel_name="rma.order.line.make.purchase.order.item",
         inverse_name="wiz_id",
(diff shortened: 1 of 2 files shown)


rma_purchase/wizards/rma_order_line_make_purchase_order.py
```diff
@@ -28,6 +33,19 @@ class RmaLineMakePurchaseOrder(models.TransientModel):
...
+        type_obj = self.env["stock.picking.type"]
+        company_id = self.env.context.get("company_id") or self.env.company.id
+        types = type_obj.search(
+            [("code", "=", "incoming"), ("warehouse_id.company_id", "=", company_id)]
+        )
+        if not types:
+            types = type_obj.search(
+                [("code", "=", "incoming"), ("warehouse_id", "=", False)]
+            )
+        return types[:1]
```
LoisRForgeFlow: @AaronHForgeFlow  Maybe you can directly call purchase.order logic here? (it is api.model)

`return self.env["purchase.order"]._get_picking_type()`

AaronHForgeFlow: Good point!

LoisRForgeFlow (approved):
LGTM

LoisRForgeFlow:
@AaronHForgeFlow could fwport when you have time?

PR #666, opened 2025-07-03
[17.0][FIX] rma_puchase: write-off differences in price for standard costs

AaronHForgeFlow:
[17.0][FIX] rma_puchase: write-off differences in price between rma line and vendor refund for standard price. For reference, see this change in Odoo standard: https://github.com/odoo/odoo/commit/a022fefbaf2f9c7041d2a790661d3cba64dfff51

The price difference account is only used for standard cost, and the price sign is already taken into account for refunds.


LoisRForgeFlow (approved):
Thanks!

PR #667, opened 2025-07-03
[16.0][FIX] rma_puchase: write-off differences in price between rma line and vendor refund for standard price.

AaronHForgeFlow:
The FWD port of: https://github.com/ForgeFlow/stock-rma/pull/403 was not entirely correct for standard price

also added a test.


LoisRForgeFlow (approved):
Thanks!

PR #673, opened 2025-07-07
[17.0][FWP][FIX] rma_sale, fix performance on search sale order lines

ArnauCForgeFlow:
Forward port of #363


LoisRForgeFlow (approved):
Works as expected :+1:

PR #678, opened 2025-07-22
[16.0][FIX] rma: RMA group view inbound vs outbound group

AaronHForgeFlow:
- The out route is in inbound --> moved to outbound
- Add the default supplier location in outbound
- The default operation has own group in the rma line form view. Here we do the same

Before:
rma_group_bad

After:
rma_group_good

### DIFF
--- a/rma/views/rma_order_view.xml
+++ b/rma/views/rma_order_view.xml
@@ -172,11 +172,15 @@
                     </group>
                 </group>
                 <group>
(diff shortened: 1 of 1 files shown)


LoisRForgeFlow (approved):
:+1:

DavidJForgeFlow (approved):

AaronHForgeFlow:
Thank you! I will do the FW Port now!

PR #681, opened 2025-07-24
[18.0][FIX] rma_repair

chafique-delli:
When we want to create a repair order from an RMA line, in the wizard, we get this error:
View types not defined tree found in act_window action undefined

@JasminSForgeFlow , @LoisRForgeFlow


LoisRForgeFlow (approved):
Thanks!

PR #682, opened 2025-08-04
[18.0][IMP] rma_repair

chafique-delli:
On the repair order, the basic field 'Under warranty' and the field 'Under warranty' linked to the RMA are confusing.
I think it's better to display only one of the two.

@JasminSForgeFlow , @LoisRForgeFlow

### DIFF
--- a/rma_repair/models/repair.py
+++ b/rma_repair/models/repair.py
@@ -10,6 +10,7 @@ class RepairOrder(models.Model):
     rma_line_id = fields.Many2one(
         comodel_name="rma.order.line", string="RMA", ondelete="restrict"
     )
-    under_warranty = fields.Boolean(
+    rma_under_warranty = fields.Boolean(
+        string="Under Warranty",
         related="rma_line_id.under_warranty",
     )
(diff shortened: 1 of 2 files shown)


rma_repair/models/repair.py
```diff
@@ -10,6 +10,7 @@ class RepairOrder(models.Model):
     rma_line_id = fields.Many2one(
         comodel_name="rma.order.line", string="RMA", ondelete="restrict"
     )
-    under_warranty = fields.Boolean(
+    rma_under_warranty = fields.Boolean(
+        string="Under Warranty",
         related="rma_line_id.under_warranty",
```
LoisRForgeFlow: This was making the repair one related to the linked RMA (at least that's the intention, I have not functionally checked). In case, we don't want to disconnect them, you should fix only the view and keep only one.

If what I describe above (have both fields related) does not work as one would expect, a refactor with proper compute methods might be needed.

LoisRForgeFlow (changes requested):
Thanks for raising the issue and contributing. However, I do not fully agree with your approach. See below.

LoisRForgeFlow:
@chafique-delli thanks for adjusting, have you tested what I mentioned?

